PR #1, opened 2019-07-18
add convexhull alg

Den3:
This PR adds Graham Scan algorithm for convex hull

### DIFF
--- a/alg_convex_hull.go
+++ b/alg_convex_hull.go
@@ -1,0 +1,98 @@
+package simplefeatures
+
+import "sort"
+
+const (
+	// Right indicates the orientation is right turn which is anticlockwise
+	Right = iota
+	// Collinear indicates three points are on the same line
+	Collinear
+	// Left indicates the orientation is left turn which is clockwise
+	Left
+)
+
+// graphamScan returns a convex hull.
+func grahamScan(ps []Point) []Point {
+	if len(ps) < 3 {
+		return nil
+	}
+	// TODO: handle the case that three points are collinear
+	sortedPoints := sortByPolarAngle(ps)
+
+	s := make([]Point, 2, len(sortedPoints))
+	copy(s, sortedPoints[:2])
+	t := make([]Point, len(sortedPoints)-2)
+	copy(t, sortedPoints[2:])
+
+	for i := 0; i < len(t); i++ {
+		if orientation(s[len(s)-2], s[len(s)-1], t[i]) == Left {
+			s = append(s, t[i])
+		} else {
+			s = s[:len(s)-1]
+			s = append(s, t[i])
+		}
+	}
+
+	return append(s, s[0])
+}
+
+func sortByPolarAngle(ps []Point) []Point {
+	ltlp := ltl(ps)
+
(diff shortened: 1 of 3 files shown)


alg_convex_hull.go
```diff
@@ -0,0 +1,119 @@
+package simplefeatures
+
+import (
+	"sort"
+)
+
+const (
+	// Right indicates the orientation is right turn which is anticlockwise
+	Right = iota
```
peterstace: Could you make `Right`, `Collinear`, and `Left` unexported? They wouldn't be needed by library users.

alg_convex_hull.go
```diff
@@ -0,0 +1,119 @@
...
+)
+
+// graphamScan returns a convex hull.
+func grahamScan(ps []XY) []XY {
+	if len(ps) < 3 {
+		return nil
+	}
+	sortedPoints := sortByPolarAngle(ps)
+
+	s := make([]XY, 2, len(sortedPoints))
```
peterstace: Could you name `s` as `stack` instead? If I understand correctly, it works like a stack.

alg_convex_hull.go
```diff
@@ -0,0 +1,119 @@
...
+			s = s[:len(s)-1]
+			s = append(s, t[i])
+		}
+	}
+
+	return append(s, s[0])
+}
+
+// soryByPolarAngle sorts the points by their polar angle
+func sortByPolarAngle(ps []XY) []XY {
```
peterstace: Since `sortByPolarAngle` works in-place, I don't think we need to return anything.

alg_convex_hull.go
```diff
@@ -0,0 +1,119 @@
...
+
+	return rpi
+}
+
+// orientation checks if s is on the right hand side or left hand side of the line formed by p and q
+// if it returns -1 which means there is an unexpected result.
+func orientation(p, q, s XY) int {
+	cp := crossProduct(p, q, s)
+	switch {
+	case cp.GT(NewScalarFromFloat64(0)):
```
peterstace: `scalar.go` defines `zero`, which you could use in place of `NewScalarFromFloat64`.

alg_convex_hull.go
```diff
@@ -0,0 +1,119 @@
...
+
+// crossProduct implements Heron's formula which returns the 2 times of area formed by p, q and s
+//         | p.X p.Y 1 |
+// 2 * S = | q.X q.Y 1 |
+//         | s.X s.Y 1 |
+// when p, q and s are clockwise, the return value is negative
+// when p, q and s are anticlockwise, the return value is positive
+func crossProduct(p, q, s XY) Scalar {
+	return q.Sub(p).Cross(s.Sub(q))
+	// return p.X.Mul(q.Y).Sub(p.Y.Mul(q.X)).Add(q.X.Mul(s.Y)).Sub(q.Y.Mul(s.X)).Mul(s.X.Mul(p.Y)).Sub(s.Y.Mul(p.X))
```
peterstace: Should be able to remove commented out code (same for `distanceSq`).

PR #2, opened 2019-07-19
WKB Marshalling

peterstace:
- Adds a new method `AsBinary(io.Writer) error` to the `Geometry interface.
- Adds implementations for `AsBinary` for each concrete geometry type.
- Adds unit tests for happy path cases for marshalling as WKB. No error handling cases have been added.


peterstace:
Thank you for reviewing @Den3 .

PR #3, opened 2019-07-19
Implement UnmarshalGeoJSON

peterstace:
- Adds `UnmarshalGeoJSON`, which takes a geojson stream and returns a `Geometry`.
- Adds happy path tests.

### DIFF
--- a/geojson.go
+++ b/geojson.go
@@ -5,17 +5,15 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"io"
 	"math"
 	"strconv"
 )
 
-func UnmarshalGeoJSON(r io.Reader) (Geometry, error) {
+func UnmarshalGeoJSON(input []byte) (Geometry, error) {
 	var firstPass struct {
 		Type string `json:"type"`
 	}
-	var rCopy bytes.Buffer
-	if err := json.NewDecoder(io.TeeReader(r, &rCopy)).Decode(&firstPass); err != nil {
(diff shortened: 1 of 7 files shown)


geojson.go
```diff
@@ -0,0 +1,175 @@
...
+	"math"
+	"strconv"
+)
+
+func UnmarshalGeoJSON(r io.Reader) (Geometry, error) {
+	var firstPass struct {
+		Type string `json:"type"`
+	}
+	var rCopy bytes.Buffer
+	if err := json.NewDecoder(io.TeeReader(r, &rCopy)).Decode(&firstPass); err != nil {
```
peterstace: The copy into the `bytes.Buffer` wouldn't be necessary if I were to pass in a `[]byte` instead of an `io.Reader`... I'll do that instead.

peterstace: Fixed.

util_test.go
```diff
@@ -20,7 +20,21 @@ func geomFromWKT(t *testing.T, wkt string) Geometry {
 func expectDeepEqual(t *testing.T, got, want interface{}) {
 	t.Helper()
 	if !reflect.DeepEqual(got, want) {
```
Den3: What do you think about https://github.com/google/go-cmp?

peterstace: That package is pretty cool, although I want to reduce the amount of dependencies as much as possible. Since we're only using `reflect.DeepEqual` for testing, I think it's ok.

PR #4, opened 2019-07-21
Convex hull integration with Geometry

peterstace:
- Integrates Graham Scan with other Geometry types (adds a `ConvexHull() Geometry` method to the `Geometry` interface).
- Adds additional tests for the convex hull algorithm.
- Changes to the convex hull algorithm so that it doesn't emit collinear points as part of the hull.


alg_convex_hull.go
```diff
@@ -1,110 +1,191 @@
...
+		return NewPoint(hull[0])
+	case 2:
+		ln, err := NewLine(
+			Coordinates{hull[0]},
+			Coordinates{hull[1]},
+		)
+		if err != nil {
+			panic("bug in grahamScan routine - output 2 coincident points")
+		}
+		return ln
```
Den3: Not quite sure why do you need to take into account the situation that `len(hull)` is less than 3 which is not a definition of convex hull.

alg_convex_hull.go
```diff
@@ -1,110 +1,191 @@
...
-		case ori == leftTurn:
-			resultStack = append(resultStack, toDoStack[i])
-		default:
-			resultStack = resultStack[:len(resultStack)-1]
-			resultStack = append(resultStack, toDoStack[i])
+	var i int // tracks progress through the ps slice
+	var stack pointStack
+	stack.push(ps[0])
+	i++
+	for i < len(ps) && len(stack) < 2 {
```
Den3: why do you not simply push the first element?

alg_convex_hull.go
```diff
@@ -1,110 +1,191 @@
...
-			return false
+		if anchor.Equals(ps[i]) {
+			return true
 		}
-		ori := orientation(virtualPoint, ps[i], ps[j])
-
-		if ori == collinear {
-			return distanceSq(virtualPoint, ps[i]).LT(distanceSq(virtualPoint, ps[j]))
+		if anchor.Equals(ps[j]) {
+			return false
```
Den3: Don't quite understand this part and above? Would you please elaborate it?

peterstace: I've added a comment, let me know if that doesn't clear it up, and I can explain further.

alg_convex_hull.go
peterstace: The first element is pushed on line 114. But then we need the `for` loop to account for the cases where there are duplicated points equal to the first point. E.g. if the points are `(3 4), (3 4), (5 5), ...`, we push point `(3 4)` on line 114, then the for loop skips over the second point (also `(3 4`), and then pushes the 3rd point `(5 5)`.

alg_convex_hull.go
peterstace: The reason is to extend the definition of a convex hull to the non-polygon cases, matching Postgis behaviour. E.g. the convex hull of a point is the same point, and the convex hull of a straight line is the same straight line.

alg_convex_hull.go
Den3: Sorry, I meant the second point.

alg_convex_hull.go
```diff
@@ -161,12 +162,17 @@ func sortByPolarAngle(ps []XY) {
 
 	ps = ps[1:] // only sort the remaining points
 	sort.Slice(ps, func(i, j int) bool {
+		// If any point is equal to the anchor point, then always put it first.
+		// This allows those duplicated points to be removed when the results
+		// stack is initiated.
```
Den3: So this one aims to remove duplicated point, but I think the duplicated point will be removed when doing graham scan, won't it?

alg_convex_hull.go
peterstace: The second point may be equal to the first point, in which case we don't want to push it (since we don't want duplicated points in the output). The third point may also be equal to the first, so we need to use a loop to get the first non-conincident point.

alg_convex_hull.go
peterstace: It's true; duplicated points will be removed by the graham scan. However, there's one edge case -- if the lowest-leftmost point is duplicated, then the naive graham scan wouldn't remove that duplication. That's because the first step of the graham scan is to grab the first two points and push them onto the result stack (but those two points might be duplicates!). By sorting all of the points coincident to the anchor point first, we can easily skip past any duplicate points when initializing the result stack.

PR #19, opened 2019-08-02
move the checking version out of graham scan algorithm

### LINKED ISSUE #18: Reduce the complexity of convex hull algorithm
Check if there is not optimisation can be used and it should also take into account any edge cases.

Den3:
Fixed #18

### DIFF
--- a/geom/alg_convex_hull.go
+++ b/geom/alg_convex_hull.go
@@ -48,108 +48,117 @@ func convexHull(g Geometry) Geometry {
 		return g
 	}
 	pts := g.convexHullPointSet()
-	hull := grahamScan(pts)
-	switch len(hull) {
+	// TODO: length may not be a good signal to indicate
+	// how to convert XY to Geometry. Need to revisit
+	switch len(pts) {
 	case 0:
 		return NewGeometryCollection(nil)
 	case 1:
-		return NewPoint(hull[0])
+		return NewPoint(pts[0])
 	case 2:
+		if pts[0].Equals(pts[1]) {
+			return NewPoint(pts[0])
+		}
 		ln, err := NewLine(
-			Coordinates{hull[0]},
-			Coordinates{hull[1]},
+			Coordinates{pts[0]},
+			Coordinates{pts[1]},
 		)
 		if err != nil {
-			panic("bug in grahamScan routine - output 2 coincident points")
+			panic(err)
 		}
 		return ln
-	default:
-		coords := [][]Coordinates{make([]Coordinates, len(hull))}
-		for i := range hull {
-			coords[0][i] = Coordinates{XY: hull[i]}
-		}
-		poly, err := NewPolygonFromCoords(coords)
-		if err != nil {
-			panic(fmt.Errorf("bug in grahamScan routine - didn't produce a valid polygon: %v", err))
-		}
-		return poly
 	}
-}
 
-type pointStack []XY
+	// TODO: check if the points are all the same
+	isSame := true
+	for i := 1; i < len(pts); i++ {
+		if !pts[0].Equals(pts[i]) {
+			isSame = false
+			break
+		}
+	}
+	if isSame {
+		return NewPoint(pts[0])
+	}
 
-func (s *pointStack) push(p XY) {
-	(*s) = append(*s, p)
-}
+	cl, ok := collinearLine(pts)
+	if ok {
+		return cl
+	}
 
-func (s *pointStack) pop() XY {
-	p := s.top()
-	(*s) = (*s)[:len(*s)-1]
-	return p
-}
-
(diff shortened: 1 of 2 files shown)


geom/alg_convex_hull.go
```diff
@@ -48,108 +48,94 @@ func convexHull(g Geometry) Geometry {
...
-	hull := grahamScan(pts)
-	switch len(hull) {
+	// TODO: length may not be a good signal to indicate
+	// how to convert XY to Geometry. Need to revisit
+	switch len(pts) {
 	case 0:
 		return NewGeometryCollection(nil)
 	case 1:
-		return NewPointXY(hull[0])
+		return NewPointXY(pts[0])
```
peterstace: The `isSame` check below seems as though it can already handle the cases where there are 1 or 2 points. So I don't think this switch is necessary (you would only have to check the 0 points case).

geom/alg_convex_hull.go
```diff
@@ -48,108 +48,94 @@ func convexHull(g Geometry) Geometry {
...
-	for i < len(ps) && len(stack) < 2 {
-		if !stack.top().Equals(ps[i]) {
-			stack.push(ps[i])
-		}
-		i++
-	}
-
-	for i < len(ps) {
-		ori := orientation(stack.underTop(), stack.top(), ps[i])
+	resultStack := make([]XY, 0, len(ps))
```
peterstace: What are your thoughts on adding back in the `pointStack` abstraction? I think having the `pop`, `push`, `top` and `underTop` abstractions really helps to make the algorithm more readable.

geom/alg_convex_hull.go
```diff
@@ -48,108 +48,94 @@ func convexHull(g Geometry) Geometry {
...
+	resultStack = append(resultStack, ps[0], ps[0])
+	toDoStack := make([]XY, len(ps)-1)
+	copy(toDoStack, ps[1:])
```
peterstace: Is there a need to copy into the `toDoStack`? If we just keep track of the position in the input `ps` array using an index, we could avoid copying the input points.

geom/alg_convex_hull.go
```diff
@@ -158,22 +144,21 @@ func sortByPolarAngle(ps []XY) {
...
 	ps = ps[1:] // only sort the remaining points
-	sort.Slice(ps, func(i, j int) bool {
-		// If any point is equal to the anchor point, then always put it first.
-		// This allows those duplicated points to be removed when the results
-		// stack is initiated.
-		if anchor.Equals(ps[i]) {
-			return true
-		}
-		if anchor.Equals(ps[j]) {
+	sort.SliceStable(ps, func(i, j int) bool {
```
peterstace: I recall that there was a problem using plain `sort.Slice` rather than `sort.SliceStable`. I'm not convinced that using a stable sort actually makes sense here, because the inputs can appear in any order (and the ordering of the inputs doesn't have any meaning). I'd prefer that we get to the bottom of _why_ `sort.Slice` wasn't working.

geom/alg_convex_hull.go
```diff
@@ -195,3 +180,36 @@ func distanceSq(p, q XY) Scalar {
 	pSubQ := p.Sub(q)
 	return pSubQ.Dot(pSubQ)
 }
+
+func collinearLine(pts []XY) (Geometry, bool) {
+	if len(pts) < 2 {
+		return nil, false
+	}
+	ps := make([]XY, len(pts))
```
peterstace: What's the reason to make a copy of the input? Even if their order is modified, I don't think that would matter for the caller.

geom/alg_convex_hull.go
```diff
@@ -195,3 +180,36 @@ func distanceSq(p, q XY) Scalar {
...
+
+	for i := 1; i < len(ps); i++ {
+		if orientation(ps[0], ps[len(ps)-1], ps[i]) != collinear {
+			return nil, false
+		}
+	}
+
+	// already check that if the initial point is as same as the end point.
+	// Ignore error here.
+	cl, _ := NewLineC(
```
peterstace: What are your thoughts on panicking instead of ignoring the error?

geom/alg_convex_hull_test.go
```diff
@@ -159,7 +159,7 @@ func TestConvexHull(t *testing.T) {
 		{
 			// there are collinear points",
 			input:  "MULTIPOINT(7 7,8 5,7 2,6 5,5 5,4 6,4 2,3 7,2 1,14 9)",
-			output: "POLYGON((2 1,7 2,14 9,7 7,3 7,2 1))",
```
peterstace: Good catch :) The test data here was definitely wrong.

image

peterstace:
As discussed, closing this one.

PR #34, opened 2019-08-16
Added hasIntersection functions and refactored orientation code.

DoctorLoki:


peterstace:
Comment mainly on the approach:

Pros:

- No geometry allocations needed for the 'intersects' calculations. And having completely different sets of functions for 'intersects' and 'intersection' algorithms will definitely allow shortcuts when it comes to more complicated cases such as polygon/polygon intersects.

Cons:

- There will be a reasonable amount of duplication between the 'intersects' and 'intersection' calculations, at least for the ones that we've implemented so far. This may not be the case for more advanced geometries, since there are more shortcuts we can take.

On the balance of things, I'm happy with the approach in general. Although I think when the following two conditions hold then we should remove the duplication (i.e. perform the intersection, and then check IsEmpty on the resulting geometry):

- If the result is always a Point or an EmptySet. These are fixed size geometries, and _probably_ would get allocated on the stack rather than the heap (although that's something to investigate). Memory allocated on the stack should have minimal impact on performance
- If the algorithm has time and memory complexity O(1), or if the algorithm has a higher upper bound complexity but there is no opportunity to early exit from the algorithm once we know that the intersection has occurred.

peterstace:
A few additional things would be good to have before merging:

1. Merge master into this branch. There will be a bunch of merge conflicts, mainly relating to https://github.com/peterstace/simplefeatures/pull/35.

2. Add some tests. There is pretty good coverage for the `Intersection` tests, so it may just be a matter of running those test cases also against the new `Intersects` method.

PR #55, opened 2019-09-04
Remove linear ring as first class geometry

peterstace:
This change removes the linear ring type. Reasons for removing it:

- I don't think it makes sense to expose the linear ring as a first class geometry. Linear rings are just building blocks that polygons are built on top of. But they don't really make sense as geometries on their own.

- Linear ring as a first class geometry is adding to the explosion of combinations needed to be handled for operations that take two geometries.

- There is no official WKT or WKB for linear rings (only unofficial extensions that some vendors support).

- The codebase as it stands before this change has a bunch of special/edge case around handling linear rings as first class geometries.

Drawbacks of removing linear ring:

- Some internal functions expect to receive a linear ring as an argument. These now take a line string. However, it's not enforced through the type system anymore that the line string is actually a linear ring (i.e. closed and simple).


peterstace:
I've self reviewed this a few times, and slept on it (because it's a somewhat big directional change rather than an incremental additional). I'm still happy with this change, but will leave it until #34 is merged so as not to introduce a merge conflict.

PR #66, opened 2019-09-22
Set up fuzz tests for intersection

peterstace:
Don't want to merge this until the intersection differences between simplefeatures and postgis have been rectified.

Most of the differences are cosmetic -- however it would still be good to get the same results so that automated testing is easier.

---
This change is [Reviewable](https://reviewable.io/reviews/peterstace/simplefeatures/66)


peterstace:
This should now be ready to review.

PR #81, opened 2019-12-28
Sweep linestring intersection

peterstace:
Benchmark change:

```
benchmark                                               old ns/op      new ns/op     delta
BenchmarkIntersectsLineStringWithLineString/n=10        1051           7877          +649.48%
BenchmarkIntersectsLineStringWithLineString/n=100       108504         88597         -18.35%
BenchmarkIntersectsLineStringWithLineString/n=1000      12071435       937373        -92.23%
BenchmarkIntersectsLineStringWithLineString/n=10000     1103327828     11241364      -98.98%
```

The breakeven-point seems to be around n=~100. However, that would only apply to the particular input case from the benchmark. The main thing that the benchmark confirms is that the complexity class has improved.

### DIFF
--- a/geom/alg_intersects.go
+++ b/geom/alg_intersects.go
@@ -1,6 +1,10 @@
 package geom
 
-import "fmt"
+import (
+	"container/heap"
+	"fmt"
+	"sort"
+)
 
 func hasIntersection(g1, g2 Geometry) bool {
 	if g2.IsEmpty() {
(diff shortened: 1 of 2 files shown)


peterstace:
There are some remaining memory allocation overheads associated with interface usage (when using `line.Intersects(other)` and also when using the `heap` data structure). I'll tackle those in a second pass. We _should_ be able to make this a zero-allocation operation.

peterstace:
Thanks for reviewing!

PR #87, opened 2020-01-05
Replace Geometry interface with sum-type

peterstace:
This change replaces the `Geometry` interface with a new `Geometry` type that can hold exactly one geometry value (i.e. it's a sum-type/variant/tagged-union). It also replaces the `AnyGeometry` type.

The new type is implemented as a pointer to the underlying geometry, along with a tag to indicate the type of geometry that it is.

Using the existing `Geometry` interface has a bunch of problems:

- It's a very large interface (the bigger the interface, the weaker the abstraction).
- Go interfaces are really designed so that new implementations that adhere to the behaviour expected by the interface can be subbed in. But that's not what we're using it for here... We typically switch on the implementation type and panic if it's not an expected type.
- With using an interface, we have nowhere to implement logic specific to all geometries, or logic that applies to multiple geometries (since you can't add methods to an interface -- the methods can only go on the types that implement the interface).

A tagged union seems like a better match to the actual use-case.


peterstace:
Note: I've self-reviewed this and am happy with the current CL.

peterstace:
Thanks for reviewing @Den3. Agreed, using the `unsafe` package is not ideal, but you're right -- it's only in a small number of places (a handful of 1 line methods restricted to a single struct). So if we for some reason need to change that part of the implementation it shouldn't be too tricky. Only using unsafe in a handful of places helps to manually verify that its usage is correct as well.

The alternative was to replace the `ptr` and `tag` fields with 9 pointers to each of the different geometry types (and only have one of them non-nil at a time). But then that might have made the `Geometry` type a little bit too heavy to pass around by value (72 bytes vs 16 bytes) -- and I'd rather have it passed around by value so that there is only a single redirection (rather than 2) when using the `Geometry` type.

Issue #84, opened 2020-01-04
Implement ST_Area for all types


peterstace:
This is to match the behaviour of PostGIS. It's currently only implemented for Polygon and MultiPolygon. Other types should give 0 (and for GeometryCollection, areas should be summed).

PR #99, opened 2020-01-17
Add Area method for all types and SignedArea for (Multi)Polygons.

DoctorLoki:
Addresses #84 by adding an Area method for geometry types.

### DIFF
--- a/geom/attr_test.go
+++ b/geom/attr_test.go
@@ -2,7 +2,6 @@ package geom_test
 
 import (
 	"math"
-	"reflect"
 	"strconv"
 	"strings"
 	"testing"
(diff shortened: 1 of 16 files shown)


geom/type_polygon.go
```diff
@@ -342,17 +342,27 @@ func (p Polygon) IsValid() bool {
...
 	n := p.NumInteriorRings()
 	for i := 0; i < n; i++ {
-		area -= areaOfLinearRing(p.InteriorRingN(i))
+		area -= math.Abs(signedAreaOfLinearRing(p.InteriorRingN(i)))
 	}
 	return area
 }
 
-func areaOfLinearRing(lr LineString) float64 {
+// SignedArea returns a positive value only for polygons with a counterclockwise winding order.
```
peterstace: The windings of the inner and outer rings could be inconsistent (e.g. both CCW or both CW). I think that's ok for this function (it would just give a weird result in that case). What are your thoughts on clarifying that behaviour in the documentation?

Maybe something like:

```
// SignedArea gives the positive area of the polygon when the outer rings are
// wound CCW and any inner rings are wound CW, and the negative area of the
// polygon when the outer rings are wound CW and any inner rings are wound CCW.
// If the windings of the inner and outer rings are the same, then the area
// will be inconsistent.
```

geom/type_geometry_collection.go
```diff
@@ -198,3 +198,16 @@ func (c GeometryCollection) IsValid() bool {
 	})
 	return all
 }
+
+// Area in the case of a GeometryCollection is the sum of the areas of its parts.
+func (c GeometryCollection) Area() float64 {
```
peterstace: Looks good.

geom/attr_test.go
```diff
@@ -476,6 +477,64 @@ func TestArea(t *testing.T) {
...
+			t.Logf("input: %s", tc.input)
+			geom := geomFromWKT(t, tc.input)
+			var got float64
+			switch {
+			case geom.IsPolygon():
+				got = geom.AsPolygon().SignedArea()
+			case geom.IsMultiPolygon():
+				got = geom.AsMultiPolygon().SignedArea()
+			default:
+				t.Errorf("expected: Polygon or MultiPolygon, got: %s", reflect.TypeOf(geom).Name())
```
peterstace: You'll just get `got: Geometry` here, which probably isn't too useful. I don't think we really have a way to actually get the name of the specific geometry type... Although we could add a method to the `Geometry` type that returns the tag name.

geom/type_sum.go
```diff
@@ -584,19 +584,29 @@ func (g Geometry) Centroid() (Point, bool) {
 }
 
 // Area gives the area of the Polygon or MultiPolygon. If the Geometry is not a
```
peterstace: Nit: Comment could be updated to indicate that it also gives the area of GeometryCollections.

peterstace:
Hmm, looks like the build is failing. I think it's because the fuzz tests need to be updated.

Issue #80, opened 2019-12-28
Implement ST_Reverse


peterstace:
https://postgis.net/docs/ST_Reverse.html

PR #100, opened 2020-01-17
Add Reverse method to all types, and a test.

DoctorLoki:
All types have a Reverse method now, addressing #80

### DIFF
--- a/geom/type_geometry_collection.go
+++ b/geom/type_geometry_collection.go
@@ -198,3 +198,17 @@ func (c GeometryCollection) IsValid() bool {
 	})
 	return all
 }
+
+// Reverse in the case of GeometryCollection reverses each component and
+// also returns them in reversed order.
+func (c GeometryCollection) Reverse() Geometry {
+	var geoms []Geometry
+	for n := 0; n < c.NumGeometries(); n++ {
+		geoms = append(geoms, c.GeometryN(n).Reverse())
+	}
+	// Reverse the geometry slice.
+	for left, right := 0, len(geoms)-1; left < right; left, right = left+1, right-1 {
+		geoms[left], geoms[right] = geoms[right], geoms[left]
+	}
+	return NewGeometryCollection(geoms).AsGeometry()
+}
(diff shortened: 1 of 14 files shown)


peterstace:
What are your thoughts on the following?

- I noticed you've got a `Geometry` type returned from the method on each concrete geometry type for the `Reverse` function. What are your thoughts on returning the specific geometry type instead? E.g. the `Reverse` method on a `Line` could return a `Line` rather than a `Geometry`. That way, the caller doesn't have to convert back to the specific type that they know the result should be.

- The downside of returning specific geometry types is that there is no common method signature that an interface could be built around. However, I don't see that as too much of a disadvantage, since the `Geometry` type is designed around handling geometries in a generic fashion (so that would be used, rather than an interface).

geom/type_line_string.go
```diff
@@ -296,3 +296,17 @@ func (s LineString) Length() float64 {
...
+// Reverse in the case of LineString outputs the coordinates in reverse order.
+func (s LineString) Reverse() Geometry {
+	coords := s.Coordinates()
+	// Reverse the slice.
+	for left, right := 0, len(coords)-1; left < right; left, right = left+1, right-1 {
+		coords[left], coords[right] = coords[right], coords[left]
+	}
+	if s, err := NewLineStringC(coords); err != nil {
+		panic("Reverse of an existing LineString should not fail")
+	} else {
```
peterstace: Nit: [Only error flow should be indented](https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow), which I would interpret to mean that there shouldn't be an else (and the `return s.AsGeometry()` should go on its own line.

Similar for MultiPolygons and Polygons below.

geom/type_geometry_collection.go
```diff
@@ -198,3 +198,17 @@ func (c GeometryCollection) IsValid() bool {
...
+
+// Reverse in the case of GeometryCollection reverses each component and
+// also returns them in reversed order.
+func (c GeometryCollection) Reverse() Geometry {
+	var geoms []Geometry
+	for n := 0; n < c.NumGeometries(); n++ {
+		geoms = append(geoms, c.GeometryN(n).Reverse())
+	}
+	// Reverse the geometry slice.
+	for left, right := 0, len(geoms)-1; left < right; left, right = left+1, right-1 {
```
peterstace: I haven't ever seen this approach for indexing when reversing a slice before, I like it :D

PR #101, opened 2020-01-19
Store only pts in line string

peterstace:
This has two main effects:

- LineString geometries with adjacent points that are coincident are now handled correctly. Previously the coincident points were discarded, which isn't correct behaviour.

- The memory usage has been reduced by 25% (we are no longer duplicating points, but now also have to track which points are duplicated).

There is a slight performance degradation, but seems to be less than 5%.

### DIFF
--- a/geom/type_line_string.go
+++ b/geom/type_line_string.go
@@ -17,26 +17,32 @@ import (
 //
 // 1. It must contain at least 2 distinct points.
 type LineString struct {
-	segments []int // index into coords
-	coords   []Coordinates
+	// coords have been deduplicated such that no two consecutive coordinates
+	// are coincident. This allows quick calculation of Line segments.
+	coords []Coordinates
+
+	// points are indexes into coords, and retain consecutive coincident
+	// points. This is so that information about the original points making up
+	// the LineString are retained.
+	points []int
 }
 
 // NewLineStringC creates a line string from the coordinates defining its
 // points.
 func NewLineStringC(pts []Coordinates, opts ...ConstructorOption) (LineString, error) {
-	segments := make([]int, 0, len(pts)-1)
-	coords := make([]Coordinates, len(pts))
-	copy(coords, pts)
+	coords := make([]Coordinates, 0, len(pts)) // may not use full capacity
+	points := make([]int, len(pts))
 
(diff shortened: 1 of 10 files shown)


geom/type_line_string.go
```diff
@@ -56,26 +51,38 @@ func (s LineString) AsGeometry() Geometry {
...
+func (s LineString) NumLines() int {
+	return len(s.segments)
+}
+
+func (s LineString) LineN(n int) Line {
+	// Line is constructed directly here, rather than via NewLineC. This is
+	// because LineN is called in a tight loop in many places, and skipping the
+	// constructor significantly speeds up the benchmarks.
+	//
+	// The two coordinates are guarenteed to not be conincident due to the way
```
DoctorLoki: Nit: Two minor typos.

geom/type_line_string.go
```diff
@@ -16,32 +17,26 @@ import (
 //
 // 1. It must contain at least 2 distinct points.
 type LineString struct {
-	lines []Line
+	segments []int // index into coords
+	coords   []Coordinates
```
DoctorLoki: Right so we wish to allow duplicated coords, but the segments will be the minimal set of lines. (I was wondering why you didn't just reduce the coords to the minimal set and deduce the lines rather than store a separate segments array, but I think that's by design.)

geom/type_line_string.go
peterstace: Thanks, fixed.

geom/type_line_string.go
peterstace: Yep, to match the behaviour of PostGIS, we need to retain duplicated coordinates. But we also want to ignore duplicated coordinates when we traverse the line segments that make up the linestring (because we've defined Lines to have non-zero length, and lots of parts in the code make use of that definition).

Another way I could have done this was to reduce the coords to the minimal set (like we had before), and then store the indexes of _duplicated_ coordinates in another slice. I didn't try out that approach -- the advantage of doing that is that the `LineN` accessor wouldn't have to perform the extra level of redirection. The downside is that the `PointN` operations would no longer be constant time (they would be logarithmic instead). I'm going to try out that approach and see how I like it.

PR #110, opened 2020-02-10
Centroid implementation

### LINKED ISSUE #83: Implement ST_Centroid for all types
It's currently only implemented for Polygon and MultiPolygon.

https://postgis.net/docs/ST_Centroid.html

DoctorLoki:
Fixes #83


geom/type_geometry_collection.go
```diff
@@ -234,3 +240,134 @@ func (c GeometryCollection) Area() float64 {
 	}
 	return sum
 }
+
+func highestDimensionIgnoreEmpties(geom Geometry) int {
```
peterstace: Nit: using `geom` as the variable name is the same as the package name, which may be a bit confusing. Thoughts on just using `g` instead? Most other places in the code use `g` as the param for a `Geometry`.

geom/type_geometry_collection.go
```diff
@@ -234,3 +240,134 @@ func (c GeometryCollection) Area() float64 {
...
+	// We could simply use Dimension() instead of this function
```
peterstace: Good comment 👍

geom/type_geometry_collection.go
```diff
@@ -234,3 +240,134 @@ func (c GeometryCollection) Area() float64 {
...
+		panic("Invalid dimensionality in centroid calculation.")
+	}
+
+	return NewPointXY(xy), valid
+}
+
+func (c GeometryCollection) computeCentroid() (highestDim, numPoints int, sumLength, sumArea, sumX, sumY float64, valid bool) {
+	n := c.NumGeometries()
+	if n == 0 {
+		panic("Invalid, empty collection\n")
```
peterstace: Should this return invalid instead? Otherwise this will panic if `Centroid` is called an a `GEOMETRYCOLLECTION EMPTY`.

geom/type_geometry_collection.go
```diff
@@ -234,3 +240,134 @@ func (c GeometryCollection) Area() float64 {
...
+		}
+		xy = XY{sumX / sumArea, sumY / sumArea}
+	default:
+		panic("Invalid dimensionality in centroid calculation.")
+	}
+
+	return NewPointXY(xy), valid
+}
+
+func (c GeometryCollection) computeCentroid() (highestDim, numPoints int, sumLength, sumArea, sumX, sumY float64, valid bool) {
```
peterstace: Nit: could use `sumXY XY` rather than `sumX, sumY float64`.

peterstace: The number of return args seems pretty high here. What are your thoughts on defining a helper struct to hold everything?

e.g.

```
type centroidCalc struct {
    highestDim int
    numPoints int
    sumXY XY
    sumLength float64
    sumArea float64
    valid bool
}
```

geom/type_line.go
```diff
@@ -172,6 +172,10 @@ func (n Line) Length() float64 {
 	return math.Sqrt(delta.Dot(delta))
 }
 
+func (n Line) Centroid() Point {
+	return NewPointF((n.a.XY.X+n.b.XY.X)/2, (n.a.XY.Y+n.b.XY.Y)/2)
```
peterstace: An alternate way to do this could be `return NewPointXY(n.a.Add(n.b.XY).Scale(0.5))`. Something to consider if you didn't see that those helpers existed (if you disagree that this makes it easier to read, then it's probably subjective so feel free to ignore my comment).

geom/type_line_string.go
```diff
@@ -305,6 +305,36 @@ func (s LineString) Length() float64 {
 	return sum
 }
 
+// Centroid gives the centroid of the coordinates of the line string.
+// It returns true iff the centroid is well defined.
+func (s LineString) Centroid() (Point, bool) {
+	if s.NumPoints() == 0 {
```
peterstace: Looks like this if statement isn't necessary (that case would be handled by `sumLength == 0`).

geom/type_line_string.go
```diff
@@ -305,6 +305,36 @@ func (s LineString) Length() float64 {
...
+		return Point{}, false
+	}
+	sumX, sumY, sumLength := centroidAndLengthOfLineString(s)
+	if sumLength == 0 {
+		return Point{}, false
+	}
+	return NewPointF(sumX/sumLength, sumY/sumLength), true
+}
+
+func centroidAndLengthOfLineString(s LineString) (sumX float64, sumY float64, sumLength float64) {
```
peterstace: Nit: could use `sum XY` rather than `sumX, sumY float64`.

geom/type_geometry_collection.go
```diff
@@ -203,11 +203,17 @@ func (c GeometryCollection) IsValid() bool {
...
+	var numEmpty int
 	for n := 0; n < c.NumGeometries(); n++ {
-		rev := c.GeometryN(n).Reverse()
-		if rev.IsEmpty() {
-			continue // Omit empty sub-geometries.
+		if c.GeometryN(n).IsEmpty() {
+			numEmpty++
 		}
+	}
+	if numEmpty == c.NumGeometries() {
```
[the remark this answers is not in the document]
DoctorLoki: Fixed.

geom/type_geometry_collection.go
DoctorLoki: Changed all instances of `geom` as a variable name in this file into `g` instead, including in other funcs.

geom/type_geometry_collection.go
DoctorLoki: Yes, fixed.

geom/type_geometry_collection.go
DoctorLoki: Done.

DoctorLoki: Done, although I've kept validity separate.

geom/type_line.go
DoctorLoki: I wish Go had operator methods to allow mathematical operations on objects like points.

geom/type_line_string.go
DoctorLoki: Fixed (removed).

geom/type_line_string.go
DoctorLoki: Done.

geom/type_geometry_collection.go
```diff
@@ -229,8 +229,140 @@ func (c GeometryCollection) Area() float64 {
...
+
+type centroidCalc struct {
+	highestDim int
+	numPoints  int
+	sumLength  float64
+	sumArea    float64
+	sumXY      XY
+}
+
+func (c GeometryCollection) sumCentroidCalc() (result centroidCalc, valid bool) {
```
peterstace: I'm not quite sure I understand why we need the `valid` flag. It seems as though an invalid centroid calculation could just be encoded naturally inside the `centroidCalc` (e.g. when the highest dim is 1, then a result of `centroidCalc{highestDim: 1, sumLength: 0}` would represent an invalid calculation).

geom/type_point.go
```diff
@@ -131,6 +131,11 @@ func (p Point) IsValid() bool {
 	return true
 }
 
+// Centroid of a point it that point.
+func (p Point) Centroid() Point {
+	return NewPointC(p.coords)
```
peterstace: Nit: could just `return p`

geom/type_geometry_collection.go
DoctorLoki: Remove validity flag.

geom/type_point.go
DoctorLoki: Done.

PR #111, opened 2020-02-11
Change the behaviour of Empty checking in Reverse for GeometryCollection

peterstace:
This is to match the behaviour of PostGIS.

I noticed this while adding some additional test cases as part of an unrelated change (the fuzz checker noticed that the behaviour didn't match).


peterstace:
Looks like this is already fixed in the centroid branch, I'll close this PR.

PR #113, opened 2020-02-15
libgeos reference implementation tests

peterstace:
- Adds reference implementation tests against libgeos (similar to the PostGIS tests).
- Changes WKT for MultiPoints to match the behaviour of libgeos rather than PostGIS.


geom/type_geometry_collection.go
```diff
@@ -199,15 +199,17 @@ func (c GeometryCollection) IsValid() bool {
...
 }
 
-// Reverse in the case of GeometryCollection reverses each component and
-// also returns them in the original order. It also omits empty components.
+// Reverse in the case of GeometryCollection reverses each component and also
+// returns them in the original order. As a special case, if the input
+// GeometryCollection has no elements or only contains empty elements, then the
+// returned GeometryCollection doesn't contain any elements.
 func (c GeometryCollection) Reverse() GeometryCollection {
+	if c.IsEmpty() {
```
peterstace: I pulled in the bugfix from the centroid branch, because some of the extra tests cases I added for the libgeos integration were failing.

Issue #85, opened 2020-01-04
Remove EmptySet type


peterstace:
EmptySet can currently be Point, LineString, or Polygon. These should just be part of their respective types.

This is better in line with the behaviour of PostGIS, and allows a nicer interface in some cases (e.g. when returning a geometry and a flag, could instead just return a flag).

peterstace:
Fixed by https://github.com/peterstace/simplefeatures/pull/117

PR #117, opened 2020-02-23
Empty type overhaul

peterstace:
- `Point`s can now be empty.
- `LineString`s can now be empty.
- `Polygon`s can now be empty.
- `EmptySet` has been removed.

Solves #85


geom/type_polygon.go
```diff
@@ -177,7 +189,7 @@ func NewPolygonC(coords [][]Coordinates, opts ...ConstructorOption) (Polygon, er
 			return Polygon{}, err
 		}
 	}
-	return NewPolygon(rings[0], rings[1:], opts...)
+	return NewPolygon(rings, opts...)
```
peterstace: This will result in a double copy of the rings. Could instead separate out the validation into a separate step (which would do no copying internally).

peterstace:
Note to self: I've finished the first pass of self-review.

PR #124, opened 2020-02-28
Remove Equals

### LINKED ISSUE #107: Remove Equals operation
It's only implemented for a few types, and is highly non-trivial for more complicated types. I think we should abandon it for now, and loop back after Intersection is complete.

peterstace:
Fixes #107


eliteGoblin:
what's the relation of EqualsExact, seems we don't want to remove it as well?

eliteGoblin:
do we want to remove `func (c Coordinates) Equals(other Coordinates) bool ` as well?

peterstace:
> what's the relation of EqualsExact, seems we don't want to remove it as well?

`EqualsExact` is to compare two geometries by checking to see if they are structurally the same. Whereas `Equals` is to check if they are spatially the same.

E.g.

| Input1 | Input2 | Equals | EqualsExact |
| --- | --- | --- | --- |
| `LINESTRING(0 0,1 1,2 2)` | `LINESTRING(0 0,2 2)` | `true` | `false` |

We already have `EqualsExact` implemented for all combinations, so no need to remove it.

peterstace:
> do we want to remove `func (c Coordinates) Equals(other Coordinates) bool ` as well?

That's a good question (it's something that I didn't consider when I was making the change).

After thinking about it for a bit, I think we _should_ remove it. It's not actually used by anything right now (i.e. it's dead code). And if we need to compare coordinates, we can just use `==`. Things may change when we inroduce 3D and Measure coordinates, but we can deal with that when we get to it.

PR #131, opened 2020-03-07
Add support for 3D and Measure (Z and M)

peterstace:
Best place to start looking if you wish to review is in `geom/type_sequence.go`. That's the fundamental new building block of all geometries, and the changes flow on from there.


peterstace:
There's a bit of inconsistency around constructor function naming at the moment (introduced by this change). To make it consistent, I'm thinking we should:

- Use `NewABCFromABC` when one geometry is constructed from other geometries (e.g. `NewMultiPointFromPoints`).
- Use `NewABC` when the geometry is constructed from a `Sequence` or a fixed number of Coordinates.
- There will still be a few special cases, but I think that will make it broadly consistent.

peterstace:
Missing tests for ForceCoordinatesType.

geom/accessor_test.go
```diff
@@ -21,12 +21,12 @@ func TestLineAccessor(t *testing.T) {
 	line := geomFromWKT(t, "LINESTRING(1 2,3 4)").AsLine()
 	t.Run("start", func(t *testing.T) {
 		got := line.StartPoint()
-		want := Coordinates{XY{1, 2}}
+		want := Coordinates{XY: XY{1, 2}, Type: DimXY}
```
DoctorLoki: Nit: The repeated use of `Coordinates{XY: XY{x, y}, Type: DimXY}` feels cumbersome, is there any way to reduce it?

fuzztests/checks.go
```diff
@@ -305,9 +308,15 @@ func CheckConvexHull(t *testing.T, want UnaryResult, g geom.Geometry) {
 	t.Run("CheckConvexHull", func(t *testing.T) {
 		got := g.ConvexHull()
 		want := want.ConvexHull
+
+		// PostGIS retains 3D and M coordinates for convex hull, which is
```
DoctorLoki: I think there needs to be a common parsed symbol to segregate PostGIS-related code. Although having notes in the comments is good, I think `if ComplyStrictlyWithPostGIS` (or something similar) would be a better way to separate and contain all code intended to deal with those differences (And a common name for such a flag would make it searchable).

geom/alg_equals_exact.go
```diff
@@ -18,17 +18,29 @@ func newEqualsExactOptionSet(opts []EqualsExactOption) equalsExactOptionSet {
 }
 
 // Tolerance modifies the behaviour of the EqualsExact method by allowing two
-// geometry control points be be considered equal if they are within the given
-// euclidean distance of each other.
+// geometry control points be be considered equal if their XY coordinates are
+// within the given euclidean distance of each other.
```
DoctorLoki: Nit: Perhaps this function should be called `XYTolerance` to make clear this is a specialised distance metric involved. Coming from a 3D background, I would find it strange that Z is not included, so making that clear in the function's name could help avoid invalid assumptions.

internal/cmprefimpl/main.go
```diff
@@ -115,3 +116,16 @@ func deduplicateGeometries(geoms []geom.Geometry) []geom.Geometry {
 	})
 	return geoms
 }
+
+// 3D and Measure geometries are skipped because the C bindings for libgeos
+// doesn't fully support Z and M value.
+func remove3DAndMeasureGeometries(geoms []geom.Geometry) []geom.Geometry {
```
DoctorLoki: Would it ever be needed to strip the Z and M coordinates rather than skip whole Z or M geometries?

geom/validation_test.go
```diff
@@ -195,18 +196,23 @@ func TestMultiPolygonValidation(t *testing.T) {
...
-			coords[0] = make([]Coordinates, sz+1)
+			floats := make([]float64, 2*(sz+1))
 			for i := 0; i < sz; i++ {
 				angle := float64(i) / float64(sz) * 2 * math.Pi
-				coords[0][i].X = math.Cos(angle)
-				coords[0][i].Y = math.Sin(angle)
+				floats[2*i+0] = math.Cos(angle)
+				floats[2*i+1] = math.Sin(angle)
+			}
+			floats[2*sz+0] = floats[0]
```
DoctorLoki: Nit: The new float sequence technique feels like a reduction in (type) safety (as compared with using more structures types). That doesn't invalidate it, but I'm just noticing the code is harder to read and feel confident about.

geom/wkb_test.go
```diff
@@ -507,14 +439,83 @@ func TestWKBMarshalValid(t *testing.T) {
...
-		t.Logf("want:\n%v", hex.Dump(want))
-		t.Logf("got:\n%v", hex.Dump(buf.Bytes()))
-		t.Errorf("mismatch")
+	for i, tt := range []struct {
+		wkt string
+		hex string
+	}{
+		{
+			"POINT EMPTY",
+			"0101000000010000000000f87f010000000000f87f",
```
DoctorLoki: Nit: It might be illustrative to break these one-liners up into multiple lines similarly to how it is done lower down for `MULTIPOINT`.

geom/marshal_unmarshal_test.go
```diff
@@ -0,0 +1,245 @@
...
+		for i, wkt := range wkts {
+			original := geomFromWKT(t, wkt)
+			if original.CoordinatesType().IsMeasured() {
+				// GeoJSON will drop Measures
+				continue
+			}
+			if original.IsEmpty() && original.CoordinatesType().Is3D() {
+				// When parsing GeoJSON, the "width" of each coordinate is used
+				// to infer XY vs XYZ geometry. If there are no coordinates,
+				// then XY is assumed.
```
DoctorLoki: Nit: Might be good to add "Therefore GeoJSON will drop empty 3D" or similar concluding thought.

geom/type_sequence_test.go
```diff
@@ -0,0 +1,182 @@
+package geom_test
+
+import (
+	"strconv"
+	"testing"
+
+	"github.com/peterstace/simplefeatures/geom"
+)
+
+func TestSequnceZeroValue(t *testing.T) {
```
DoctorLoki: Typo.

geom/type_sequence.go
```diff
@@ -0,0 +1,145 @@
...
+// getLine extracts a 2D line segment from a sequence by joining together
+// adjacent locations in the sequence. It is designed to be called with i equal
+// to each index in the sequence (from 0 to n-1). The flag indicates if the
+// returned line is valid.
+func getLine(seq Sequence, i int) (Line, bool) {
+	if i == 0 {
+		return Line{}, false
+	}
+	ln := Line{
+		a: Coordinates{Type: DimXY, XY: seq.GetXY(i - 1)},
```
DoctorLoki: Does this usage match the function comment?

geom/geojson_unmarshal.go
```diff
@@ -0,0 +1,333 @@
...
+		// The GeoJSON spec allows parsers to ignore any "extra" coordinate
+		// values in addition to the normal 3 coordinate values used to specify
+		// a 3D position. So having a length strictly greater than 3 is not an
+		// error.
+		if length >= 3 {
+			has3D = true
+		}
+	}
+	ctype := DimXY
+	if !has2D && has3D {
```
DoctorLoki: Nice!

geom/accessor_test.go
peterstace: Good point, I've now created a helper, which removes some of the clutter.

geom/alg_equals_exact.go
peterstace: Agreed, renamed (we'll also want to add tolerance functions for Z and M, I'll created tickets for those).

internal/cmprefimpl/main.go
peterstace: That's a good point. If we strip Z and M instead of discarding the entire geometry, we will end up with more test cases which is always a good thing. Fixed.

geom/validation_test.go
peterstace: Agreed, there are two aspects where we lose type safety:

- These is less safely in the size of the floats slice. There is nothing to enforce that the length is a multiple of the coordinate type dimension. The code (purposefully and in a documented manner) panics in this case. Even though it's documented, that's definitely a con.
- It's less clear explicit which elements of the slice are Xs and Ys (and Ms and Zs).

In some sense, it does _improve_ type safety (for some aspects of the code). The main alternative that I considered was to store a slice of `Coordinates` rather than a slice of `float64`s. If we were to continue to keep track of the coordinates type within the `Coordinates` type, then these could potentially differ within a sequence (and we would have to check for that and give an error). This is actually similar to the way that rings work in polygons in this change (which I'm not too happy about...). The other more obvious disadvantage to that approach is that we're storing Z and M values even if they're not used.

geom/wkb_test.go
peterstace: Good call. WKBs are hard enough to read at the best of times. Fixed (note that I've left the WKBs one single lines for the test at the top of the file, since I autogenerated those ones rather than writing them by hand).

geom/type_sequence.go
peterstace: I think it does. The idea is that the caller should call the function with `i` equal to each index of the sequence. The trouble is that there are `n-1` lines (or less in the case of duplicated points) in a sequence. So some of the values of `i` will have to return `false`. I wanted to leave that as an implementation detail of this function that the caller shouldn't have to worry about. That way we're free to change the implementation if there's a good reason to without having to change the callers (which we might have to do if the contract of the function was that the caller shouldn't call the function with `i=0`).

geom/type_sequence_test.go
peterstace: Fixed.

geom/marshal_unmarshal_test.go
peterstace: Good call, I've now elaborated on that comment.

fuzztests/checks.go
peterstace: In this case, the behaviour here is more to work around the differences so that we don't get a false negative test failure. I'm less keen to add "compatibility mode" flags (I think that's what you're referring to?) since it adds a bit of complexity, and I'm not sure right now what the usecase for differing compatibility would be (postgis vs libgeos vs potentially others).

geom/type_sequence.go
DoctorLoki: OK, I just wanted to ensure you were being explicitly vague about whether `i==0` or `i==1` yields the first line. And as you say, duplicates could make the mapping even less clear. I'm just too used to seeing zero meaning the first element, and of course in a C/C++ world where the number of elements in an array might be construed as a `size_t` or some other `unsigned` type, the use of `i-1` looked suspicious (but is actually perfectly fine here).

PR #138, opened 2020-03-27
Wrap GEOS DE-9IM spatial predicates

peterstace:
- Creates a new package which is a cgo wrapper around libgeos.
- Currently this implements all named spatial predicates _except for_ Crosses and Overlaps (these are a bit harder than the rest, because they need dynamic DE-9IM masks based on the inputs, rather than static masks).
- Libgeos methods that return geometries are out of scope for this PR (e.g. Union and Intersection).

To properly review this PR, you will have to understand the DE-9IM model (or at least have a high level idea of what it's all about). I would recommend skimming https://en.wikipedia.org/wiki/DE-9IM before attempting to review.


libgeos/libgeos.go
```diff
@@ -0,0 +1,145 @@
...
+
+	"github.com/peterstace/simplefeatures/geom"
+)
+
+var (
+	globalMutex  sync.Mutex
+	globalHandle *Handle
+)
+
+func getGlobalHandle() (*Handle, error) {
```
DoctorLoki: I think this function guarantees if there is no error that the handle will be non-nil?

libgeos/libgeos_handle.go
```diff
@@ -0,0 +1,236 @@
...
+
+/*
+#cgo linux CFLAGS: -I/usr/include
+#cgo linux LDFLAGS: -L/usr/lib -lgeos_c
+#include "geos_c.h"
+#include <stdlib.h>
+#include <string.h>
+
+void sf_error_handler(const char *message, void *userdata) {
+	strncpy(userdata, message, 1024);
```
DoctorLoki: Note, `strncpy` writes 1024 bytes to `userdata`, and if the `message` is not '\0' terminated within its first 1024 bytes, `strncpy` will not write a terminating '\0' byte. You may wish to ensure the final byte is a '\0' byte by setting `userdata[1023] = '\0'` yourself manually, depending on whether it is required that it be a '\0' terminated string.

libgeos/libgeos_handle.go
```diff
@@ -0,0 +1,236 @@
...
+	"fmt"
+	"strings"
+	"unsafe"
+
+	"github.com/peterstace/simplefeatures/geom"
+)
+
+// Handle is an opaque handle that can be used to invoke libgeos operations.
+// Instances are not threadsafe and thus should not be used in a concurrent
+// manner.
```
DoctorLoki: I think it would be better to write this comment as "Instances are not threadsafe and thus should only be used within some mutex-protected manner."

libgeos/libgeos_handle.go
```diff
@@ -0,0 +1,236 @@
...
+// always returns a non-nil error. If no error message has been reported, then
+// it returns a generic error message.
+func (h *Handle) err() error {
+	msg := h.errMsg()
+	if msg == "" {
+		// No error stored, which indicates that the error handler didn't get
+		// trigged. The best we can do is give a generic error.
+		msg = "libgeos internal error"
+	}
+	h.errBuf = [1024]byte{} // Reset the buffer for the next error message.
```
DoctorLoki: Comment: I was thinking there might be a `memset`-like way to zero the array that is more idiomatic, but apparently not.

libgeos/libgeos_handle.go
```diff
@@ -0,0 +1,236 @@
...
+}
+
+// errMsg gets the textual representation of the last error message reported by
+// libgeos.
+func (h *Handle) errMsg() string {
+	// The error message is either NULL terminated, or fills the entire buffer.
+	firstZero := len(h.errBuf)
+	for i, b := range h.errBuf {
+		if b == 0 {
+			firstZero = i
```
DoctorLoki: Or you could omit the `firstZero` variable entirely and return `string(h.errBuf[:i])` here and return `string(h.errBuf)` later.

libgeos/libgeos_handle.go
```diff
@@ -0,0 +1,236 @@
...
+			break
+		}
+	}
+	return string(h.errBuf[:firstZero])
+}
+
+// Release releases any resources held by the handle. The handle should not be
+// used after Release is called.
+func (h *Handle) Release() {
+	C.GEOS_finish_r(h.context)
```
DoctorLoki: Comment: Some C/C++ practioners advocate zeroing pointers such as `h.context` in cases like this to force an error if it is used after being released.

libgeos/libgeos_handle.go
```diff
@@ -0,0 +1,236 @@
...
+	defer C.GEOSGeom_destroy(gh1)
+
+	gh2, err := h.createGeometryHandle(g2)
+	if err != nil {
+		return false, err
+	}
+
+	defer C.GEOSGeom_destroy(gh2)
+	cmask := C.CString(mask)
+	defer C.free(unsafe.Pointer(cmask))
```
DoctorLoki: Since these `mask` strings are a known fixed length, it'd be faster to avoid `malloc` and `free` entirely by using fixed-sized arrays of 10 bytes. Also, it looks like the code here doesn't check for the case that creating the C string fails?

libgeos/libgeos.go
```diff
@@ -0,0 +1,145 @@
...
+}
+
+// Contains returns true if and only if geometry A contains geometry B.
+// See the global Contains function for details.
+// Formally, the following two conditions must hold:
+//
+// 1. No points of B lies on the exterior of geometry A. That is, B must only be
+// in the exterior or boundary of A.
+//
+// 2.At least one point of the interior of B lies on the interior of A. That
```
DoctorLoki: Nit: Spacing.

libgeos/libgeos.go
peterstace: Yes, that's correct.

libgeos/libgeos_handle.go
peterstace: Yep, that makes sense.

This should be ok, since we don't require that `userdata` is a null terminated string. It's a pointer to a `[1024]byte` chunk of Go memory, and we don't treat it as necessarily being null terminated when we extract a Go string out of it.

libgeos/libgeos_handle.go
peterstace: Agreed, that's a bit clearer. I've switched the wording to say how it _should_ be used rather than how it _shouldn't_ be used (I didn't use your exact wording, but I think that's the distinction you were getting at).

libgeos/libgeos_handle.go
peterstace: Yep, seems like this is the way to do it (btw, the compiler will complain at you if you don't use the correct size, since the size makes part of the type -- so that's sort of nice).

libgeos/libgeos_handle.go
peterstace: Good suggestion, that's a bit easier to understand. Fixed.

libgeos/libgeos_handle.go
```diff
@@ -0,0 +1,236 @@
+package libgeos
+
+/*
+#cgo linux CFLAGS: -I/usr/include
```
Den3: This will be an issue for Mac Catalina because App has remove this folder entirely, so may be good to have an issue for that?

libgeos/libgeos_handle.go
```diff
@@ -0,0 +1,236 @@
...
+
+// err gets the last error message reported by libgeos as an error type. It
+// always returns a non-nil error. If no error message has been reported, then
+// it returns a generic error message.
+func (h *Handle) err() error {
+	msg := h.errMsg()
+	if msg == "" {
+		// No error stored, which indicates that the error handler didn't get
+		// trigged. The best we can do is give a generic error.
+		msg = "libgeos internal error"
```
Den3: Just wondering why does it always return error?

libgeos/libgeos_handle.go
```diff
@@ -0,0 +1,236 @@
...
+var ErrGeometryCollectionNotSupported = errors.New("GeometryCollection not supported")
+
+// Equals returns true if and only if the input geometries are spatially equal.
+func (h *Handle) Equals(g1, g2 geom.Geometry) (bool, error) {
+	if g1.IsEmpty() && g2.IsEmpty() {
+		// Part of the mask is 'dim(I(a) ∩ I(b)) = T'.  If both inputs are
+		// empty, then their interiors will be empty, and thus
+		// 'dim(I(a) ∩ I(b) = F'. However, we want to return 'true' for this
+		// case. So we just return true manually rather than using DE-9IM.
+		return true, nil
```
Den3: Is it doing this because libgeos does the same way?

libgeos/libgeos_handle.go
```diff
@@ -0,0 +1,236 @@
...
+	defer C.GEOSGeom_destroy(gh2)
+	cmask := C.CString(mask)
+	defer C.free(unsafe.Pointer(cmask))
+
+	switch ret := C.GEOSRelatePattern_r(h.context, gh1, gh2, cmask); ret {
+	case 2:
+		return false, h.err()
+	case 1:
+		return true, nil
+	case 0:
```
Den3: Should it use `const` name for the value instead of just value itself. Just checked the source code. What I could understand is that 2 means something goes wrong, 1 means ture and 0 means false. But that would be good to have a string name instead of value to help people to read it

libgeos/libgeos_handle.go
peterstace: Sounds good, that will prevent a use-after-free error. Done.

libgeos/libgeos_handle.go
peterstace: That's a good point, fixed.

libgeos/libgeos.go
peterstace: Fixed.

libgeos/libgeos_handle.go
peterstace: Good suggestion, thanks! Fixed.

peterstace: Good suggestion, thanks! Fixed.

libgeos/libgeos_handle.go
peterstace: The `err()` function should only be called if something has gone wrong with an interaction with libgeos (as reported via the return code from the libgeos call). So we always want to give back some sort of error. But there could theoretically be a case where libgeos reports an error via its return value, but _didn't_ call the `sf_error_handler` callback (in which case we won't have an error message). Setting `msg` to a generic string is just so that the user gets some sort of error message (even though the message would not be very useful).

libgeos/libgeos_handle.go
peterstace: Yes, that's correct. We have to do this because `Equals` and `Relate("T*F**FFF*")` aren't _exactly_ the same. Libgeos _does_ do this sort of logic for its `Equals` method as well (and Postgis does something very similar for `ST_Equals` as well).

I _could have_ wrapped libgeos `Equals` method instead of just wrapping the `Relate` method. However, my idea is that if we only wrap the `Relate` method, then we can later switch it out once we build the DE-9IM engine in Go.

libgeos/libgeos_handle.go
peterstace: Hmm, this seems like a major problem... I'm ok with this not working on Windows, but it seems like a major problem if it doesn't work on Mac. I'll find a solution for this before merging.

peterstace: I think I've found a solution. Thanks for pointing out this problem :)

PR #171, opened 2020-04-15
Use r-tree for LineString's IsSimple

peterstace:
This fixes the "worst case scenario" for this algorithm, which occurs
when the line segments in the LineString are stacked on top of each
other vertically (vertical zig zag).

Performance change:

```
benchmark                                     old ns/op     new ns/op     delta
BenchmarkLineStringIsSimpleZigZag/10-4        1499          3683          +145.70%
BenchmarkLineStringIsSimpleZigZag/100-4       72693         93119         +28.10%
BenchmarkLineStringIsSimpleZigZag/1000-4      5991268       1265259       -78.88%
BenchmarkLineStringIsSimpleZigZag/10000-4     585784268     16355471      -97.21%
```

Worse for smaller inputs, but quickly catches up for inputs > 100. Significantly better complexity class (quadratic vs linearithmic). I suspect there are some "large" constants for the new solution related to memory management in the rtree.

### DIFF
--- a/geom/type_line_string.go
+++ b/geom/type_line_string.go
@@ -96,6 +96,12 @@ func (s LineString) IsSimple() bool {
 		return true
 	}
 
+	// We need to track the index of the previous line segments, so that we can
+	// ignore the case where adjacent line segments intersect at a point (due
+	// to their construction). We can just subtract 1 from the current index,
+	// since there could be duplicated vertices in the middle of the sequence.
+	prev := -1
+
 	var tree rtree.RTree
 	n := s.seq.Length()
 	for i := 0; i < n; i++ {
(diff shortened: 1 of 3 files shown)


geom/type_line_string.go
```diff
@@ -90,79 +91,73 @@ func (s LineString) appendWKTBody(dst []byte) []byte {
...
 			// The first and last segment are allowed to intersect at a point,
 			// so long as that point is the start of the first segment and the
 			// end of the last segment (i.e. the line string is closed).
-			if (current == 0 && other == n-1) || (current == n-1 && other == 0) {
-				if s.IsClosed() {
-					continue
-				} else {
-					return false
-				}
+			atLoop := (i == first && j == last) || (i == last && j == first)
```
peterstace, line 145: Because of the way we iterate through `i`, would should be able to simplify this condition (remove one half of the OR).

PR #176, opened 2020-04-19
Use uint64 rather than int for r-tree record ID

peterstace:
By using a uint64, we open up the possibility do do things other than
just storing an index to an array with the record id. E.g. we could
treat the uint64 as two uint32s, and store two different indexes. This
would be less clean with an int, because we can't be 100% guarenteed
that we have 64 bits to play with.

### DIFF
--- a/rtree/rtree.go
+++ b/rtree/rtree.go
@@ -14,11 +14,9 @@ type node struct {
 
 // entry is an entry under a node, leading either to terminal items, or more nodes.
 type entry struct {
-	box Box
-
-	// For leaf nodes, this is a record ID. For non-leaf nodes, this is the
-	// index of the child node.
-	index int
+	box      Box
+	child    int    // Only populated for non-leaf nodes.
+	recordID uint64 // Only populated for leaf nodes.
 }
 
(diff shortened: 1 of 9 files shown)


rtree/rtree.go
```diff
@@ -14,11 +14,9 @@ type node struct {
...
 // entry is an entry under a node, leading either to terminal items, or more nodes.
 type entry struct {
-	box Box
-
-	// For leaf nodes, this is a record ID. For non-leaf nodes, this is the
-	// index of the child node.
-	index int
+	box      Box
+	child    int    // Only populated for non-leaf nodes.
+	recordID uint64 // Only populated for leaf nodes.
```
peterstace, line 19: `child` and `recordID` could _technically_ be stored in the same variable. However, I'm going to come back and revisit the internal structure of the rtree anyway (there are a few low level efficiency gains to be had, e.g. using arrays rather than slices).

peterstace:
Abandoning this for now. There isn't really a concrete usecase for it, and it introduces type coversions wherever the rtree is used (too much added complexity at no benefit).

If we need it in the future we can always come back to this.

PR #179, opened 2020-04-24
Improve performance of MultiPolygon validation when multiple lines touch

peterstace:
This improves the performance of MultiPolygon validation in the case where there are many touching points between the child Polygons.

The added benchmark tests uses polygons with this exact case. E.g. (with n=5):

image

In the previous algorithm, we were performing many "point in ring" operations, which are linear in the size of the ring. This change basically indexes the rings, which speeds up the point-in-ring operations.


geom/alg_intersection.go
```diff
@@ -2,16 +2,34 @@ package geom
...
-func intersectionOfMultiLineStringAndMultiLineString(
-	mls1, mls2 MultiLineString,
-) (MultiPoint, MultiLineString) {
-	return intersectionOfLines(mls1.asLines(), mls2.asLines())
+func intersectionOfIndexedLines(
+	lines1 indexedLines, lines2 indexedLines,
+) (
+	MultiPoint, MultiLineString,
+) {
+	// TODO: Investigate potential speed up of swapping lines.
```
peterstace, line 10: I'll create an additional ticket for this. To work out if it's worthwhile, we'll need to add some benchmarks specifically around this.

geom/alg_point_in_ring.go
```diff
@@ -10,9 +14,9 @@ const (
 	exterior side = +1
 )
 
-// pointRingSide checks the side of a ring that a point is on. It assumes that
+// relatePointToRing checks the side of a ring that a point is on. It assumes that
```
peterstace, line 17: We probably want to replace the implementation of this with a [winding number](https://en.wikipedia.org/wiki/Winding_number) algorithm instead. While it would have the same time complexity, it should have a much smaller constant (which can be verified using a benchmark). I'll create a ticket for that.

geom/alg_point_in_ring.go
```diff
@@ -42,25 +46,63 @@ func pointRingSide(pt XY, ring LineString) side {
...
+	})
+	if onBoundary {
+		return boundary
+	}
+
+	extent, ok := polyBoundary.tree.Extent()
+	if !ok {
+		return exterior
 	}
+	ray := line{pt, XY{extent.MaxX + 1, pt.Y}}
```
peterstace, line 96: We could be a bit smarter about how we construct the ray. We should construct the ray such that it is the shortest possible distance from the `pt` to the outside of `extent`. This could be as simple as checking the lengths of the rays constructed by going from the pt to the outside of the envelope in the 4 cardinal directions. I'll create an additional ticket for this.

geom/rtree.go
```diff
@@ -0,0 +1,32 @@
+package geom
+
+import "github.com/peterstace/simplefeatures/rtree"
+
+// TODO: make this a method on Envelope
```
peterstace, line 5: I'll make a ticket for this.

geom/type_multi_polygon.go
```diff
@@ -103,101 +121,57 @@ func validateMultiPolygon(polys []Polygon, opts ...ConstructorOption) error {
...
-				}
+func validatePolyNotInsidePoly(p1, p2 indexedLines) error {
+	// For each point where the boundaries of the two polygons intersect, we
+	// take points clockwise and counterclockwise along the boundaries and see
+	// if they are inside the opposing polygon. If they are, then the interiors
+	// intersect.
+
+	for j := range p2.lines {
+		// Find intersection points.
+		pts := make(map[XY]struct{}) // TODO: Could reuse map
```
peterstace, line 132: We don't actually need to use a map here. Instead, we could just store the points in the slice, sort the points, and then do a uniquify pass over the slice.

geom/rtree.go
peterstace: Fixed in 056e254

geom/alg_point_in_ring.go
```diff
@@ -42,25 +46,63 @@ func pointRingSide(pt XY, ring LineString) side {
...
-				count++
-			}
-		} else {
-			count++
+		return otherY < pt.Y
+	}
+	return true
+}
+
+func relatePointToPolygon(pt XY, polyBoundary indexedLines) side {
```
peterstace, line 77: Need to add some additional tests cases around this (could potentially modify the existing test suite for the non-indexed version of this algorithm).

peterstace: Additional tests added in 55c7cd9

geom/type_multi_polygon.go
peterstace: Fixed in 6c638bf

Issue #238, opened 2020-07-12
Add AreaWithTransform calculation


peterstace:
It would be useful to have an `AreaWithTransform` function that accepts a `func(XY) XY` argument.

Currently, we transform the geometry, and then take the area. But we could do this via inlining the transformation meaning that we don't have to allocate a whole new geometry.

peterstace:
Fixed via #239

PR #239, opened 2020-07-22
Add Area calculation options (signed and transform)

peterstace:
## Description

It's useful to calculate the area of a geometry under a transformation. This is possible without this change. The user would have to first call `TransformXY`, and then call `Area`. This has two problems though:

- `TransformXY` gives back an error that the user doesn't really care about. They could add the `DisableAllValidations` option when transforming, but they still have the error returned that they should check.
- Transforming creates a whole new geometry, which would only be needed for the area calculation. This is a bunch of unnecessary allocations.

This PR instead adds options to the `Area` methods, that can specify a transformation. The transformation can be done on the fly, which solves both problems.

I also refactored the "signed" area calculations into the same options framework. Signed areas and transformed areas are orthogonal, so should be able to be used together in any combination.

## Check List

Have you:

- Added unit tests? Yes

## Related Issue

- #238 

## Benchmark Results

- TODO (not expecting any change)

### DIFF
--- a/geom/attr_test.go
+++ b/geom/attr_test.go
@@ -613,21 +613,36 @@ func TestSignedArea(t *testing.T) {
 		t.Run(tc.name, func(t *testing.T) {
 			t.Logf("input: %s", tc.input)
 			geom := geomFromWKT(t, tc.input)
-			var got float64
-			switch {
-			case geom.IsPolygon():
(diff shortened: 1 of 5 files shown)


peterstace:
Thanks for reviewing :)

Issue #237, opened 2020-07-09
Add Winding Order calculation


peterstace:
We want a way to calculate the winding order of linear rings. The motivating usecase of this is as a way to detect the need to re-orient a linear ring for inclusion in a shapefile. The API design should be agnostic of that though, and applicable to more general scenarios.

A bit of thought needs to go into the best way to design the API for this. An obvious first attempt at an API design would be:

```
func IsWindingOrderClockwise(seq Sequence) bool {
    ...
}
```

This raises some questions of what to do in the following scenarios though:

- What if there aren't enough points? We need at least 3 points to reasonably define winding order. Should we return false if there aren't enough points?
- What if the input Sequence isn't actually a LinearRing? E.g. start and endpoints aren't the same, or contains self intersections.

Self intersection checks can be expensive, and may have already been performed (or the input may otherwise have been determined to be a valid linear ring).

If the function was instead:

```
func IsOuterRingWindingOrderClockwise(poly Polygon) bool {
    ...
}
```

Then we're in a better assumption to assume that the outer linear ring is actually a linear ring (since constructed polygons are assumed to have passed validity checks). But that signature doesn't fit our motivating usecase too well, since we would also want to re-orient _inner_ rings as well.

A bit more thought needs to go into the API design before we can address this.

peterstace:
Another alternative is to go with the approach of passing in a `Sequence`, but then documenting that if the user passes in a sequence that self intersects, then the output isn't well defined.

```
// IsWindingOrderClockwise ... blah blah ... if you pass in a self interesting sequence then don't rely on the output having any particular meaning ...
func IsWindingOrderClockwise(seq Sequence) bool {
    ...
}
```

peterstace:
We may be hitting into the [xy problem](http://xyproblem.info/) here...

The problem really is that we want to orient polygons and multipolygons as appropriate for shapefiles.

So we could just add methods to do that. E.g.

```
// ForceRHR returns the version of the Polygon that
// follows the right hand rule (i.e. the area that is bounded
// by the polygon is to the right of the boundary).
func (p Polygon) ForceRHR() Polygon {
    ....
}
```

peterstace:
Fixed via #240

PR #240, opened 2020-07-24
Force Ring Orientation

peterstace:
## Description

Allows to force either CW or CCW orientation.

## Check List

Have you:

- Added unit tests? Yes

- Add cmprefimpl tests? (if appropriate?) Yes

## Related Issue

- #237 

## Benchmark Results

- N/A - new feature.

### DIFF
--- a/geom/attr_test.go
+++ b/geom/attr_test.go
@@ -1115,30 +1115,42 @@ func TestForceWindingDirection(t *testing.T) {
 						(1 1,1 2,2 2,2 1,1 1),
 						(3 1,3 2,4 2,4 1,3 1))`,
 		},
-		//{
-		//	desc:     "multipolygon with single poly wound CW",
-		//	input:    "",
-		//	forceCW:  "",
-		//	forceCCW: "",
-		//},
-		//{
-		//	desc:     "multipolygon with single poly wound CCW",
-		//	input:    "",
-		//	forceCW:  "",
-		//	forceCCW: "",
-		//},
-		//{
-		//	desc:     "multipolygon with two polys of mixed winding",
-		//	input:    "",
-		//	forceCW:  "",
-		//	forceCCW: "",
-		//},
-		//{
-		//	desc:     "geometry collection containing poly and multipoly",
-		//	input:    "",
-		//	forceCW:  "",
-		//	forceCCW: "",
-		//},
+		{
+			desc:     "multipolygon with single poly wound CW",
+			input:    "MULTIPOLYGON(((0 0,0 1,1 1,1 0,0 0)))",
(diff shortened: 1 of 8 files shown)


geom/type_polygon.go
```diff
@@ -495,3 +495,30 @@ func (p Polygon) PointOnSurface() Point {
...
+// counter-clockwise orientation and any inner rings in a clockwise
+// orientation.
+func (p Polygon) ForceCCW() Polygon {
+	return p.forceOrientation(false)
+}
+
+func (p Polygon) forceOrientation(forceCW bool) Polygon {
+	orientedRings := make([]LineString, len(p.rings))
+	for i, ring := range p.rings {
+		alreadyCW := signedAreaOfLinearRing(ring, nil) < 0
```
DoctorLoki, line 516: Can you comment about whether it'd be worthwhile checking if the entire polygon is already oriented correctly and if so return that?

geom/attr_test.go
```diff
@@ -1035,3 +1035,134 @@ func TestForceCoordinatesType(t *testing.T) {
...
+			input:    "POLYGON((0 0,4 0,0 4,0 0),(1 1,1 2,2 1,1 1))",
+			forceCW:  "POLYGON((0 0,0 4,4 0,0 0),(1 1,2 1,1 2,1 1))",
+			forceCCW: "POLYGON((0 0,4 0,0 4,0 0),(1 1,1 2,2 1,1 1))",
+		},
+		{
+			desc:     "polygon with outer ring wound CCW and inner ring wound CCW",
+			input:    "POLYGON((0 0,4 0,0 4,0 0),(1 1,2 1,1 2,1 1))",
+			forceCW:  "POLYGON((0 0,0 4,4 0,0 0),(1 1,2 1,1 2,1 1))",
+			forceCCW: "POLYGON((0 0,4 0,0 4,0 0),(1 1,1 2,2 1,1 1))",
+		},
```
DoctorLoki, line 1087: Good test cases.

geom/type_polygon.go
peterstace: That's a good point. We would save additional allocations that way, since if the polygon is already oriented correctly then we don't need to allocate a new rings slice.

I'll add in some benchmarks as well, since this would technically fall into a performance optimisation.

DoctorLoki: Yes, you're right, it's a performance optimisation so I think it needn't be in the same PR. Could this kind of change also apply to other geometry types?

peterstace: Yep, probably best to do this as a different change. I've made a ticket here: https://github.com/peterstace/simplefeatures/issues/241

Yep, would apply to MultiPolygon and GeometryCollection as well.

peterstace:
Thanks for reviewing!

Issue #275, opened 2020-10-21
Investigate possible bug in validation algorithm


peterstace:
GEOS has emitted the attached WKBs, and simplefeatures as reporting them as invalid. There are two possibilities:

- GEOS is emitting invalid WKBs
- Simplefeatures is giving a false positive for detecting invalid geometries.

The task here is to investigate to see which of the two cases we are looking at.
[possible_validation_bug.txt](https://github.com/peterstace/simplefeatures/files/5413075/possible_validation_bug.txt)

peterstace:
The first multipolygon in the attached file contains 6 child polygons. The 4th child polygon's exterior ring is not simple, which violates the validation rules for polygons. In particular, the offending control points in the exterior ring are:

```
Index  Point
297    {151.1567491196036 -33.578014746144}
298    {151.15686002573   -33.57427243628813}
299    {151.15686002573   -33.57427243628814}
300    {151.15686002573   -33.57427243628813}
301    {151.1568730920001 -33.57428004599996}
302    {151.1569620060001 -33.57433404699998}
```

The 298th and 300th control point are _identical_, so this ring doubles back on itself (thus making it non-simple and thus the Polgyon invalid).

**This confirms that GEOS has emitted an invalid geometry it this case, and simplefeatures' MultiPolygon validation algorithm is correctly identifying it as such**.

Interestingly the 299th control point is only a single ULP away from the 298th and 300th control point, so we're dealing with a very marginal case here (so it's understandable that GEOS doesn't get this 100% correct).

---

What I'm thinking right now is that we probably have to just deal with GEOS sometimes giving invalid geometries. There are a few possible ways to work around this:

- Require that the user of simplefeatures provide the `DisableAllValidations` when using GEOS operations such as `geos.Intersection`, ` geos.Union` etc. This has the disadvantage that the user now has an invalid geometry, and thus has to be very careful what they actually do with it.

- Before giving back the geometry to the user, run the geometry through `extern GEOSGeometry GEOS_DLL *GEOSMakeValid_r(GEOSContextHandle_t handle, const GEOSGeometry* g);`. This has the advantage that the user should end up with a valid geometry, but it has the disadvantage that everything will be a tiny bit slower due to the extra processing.

I'm thinking that option 2 is probably the best.

peterstace:
[...]

PR #276, opened 2020-10-21
Pass results of GEOS binary ops through GEOSMakeValid_r

peterstace:
## Description

This is because GEOS sometimes creates invalid geometries. If we can
easily fix them, we should do so before passing them back to users.

## Check List

Have you:

- Added unit tests? No, I wasn't able to reproduce this in a unit test easily. It's fairly simple though.

## Related Issue

- Relates to https://github.com/peterstace/simplefeatures/issues/275

## Benchmark Results

- N/A

### DIFF
--- a/geos/libgeos.go
+++ b/geos/libgeos.go
@@ -577,7 +577,16 @@ func binaryOperation(
 	}
 	defer C.GEOSGeom_destroy(resultGH)
 
-	return h.decode(resultGH, opts)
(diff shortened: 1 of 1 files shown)


peterstace:
Seems like `GEOSMakeValid_r` isn't defined in GEOS 3.7.1 (it's only introduced in GEOS 3.8.0). This might not be a good idea to merge, since it would restrict the GEOS version it can work with.

Issue #254, opened 2020-08-29
Implement Union/Intersection/Difference using DCEL


peterstace:

peterstace:
Outstanding tasks:

- [ ] Support GeometryCollection inputs.
- [ ] Use acceleration structure for re-ordering edges around vertices.
- [ ] Use vector operations for radial ordering in `fixVertex` (rather than `math.Atan2`).
- [ ] Rethink `signedHorizontalDistanceBetweenXYAndLine` implementation.
- [ ] Use state-of-art disjoint set algorithm rather than linear/naive algorithm.
- [ ] Better linear output generation (currently emits single line segments in a MultiLineString).
- [ ] Replace all panics with error handling.
- [ ] Insert verticies from input geometries into nodeset before performing re-noding.
- [ ] Add a benchmark at least for the basic Polygon/Polygon intersection case.

peterstace:
I've created issues for the remaining tasks.

PR #278, opened 2020-10-30
DCEL - Union, Intersection, Difference, Symmetric Difference

peterstace:
## Description

Adds operations that require [DCELs](https://en.wikipedia.org/wiki/Doubly_connected_edge_list) -- Union, Intersection, Difference, and Symmetric Difference.

It's possible to build other operations on top of DCELs, but these haven't been implemented yet.

## Check List

Have you:

- Added unit tests? Yes

- Add cmprefimpl tests? (if appropriate?) Yes

## Related Issue

- https://github.com/peterstace/simplefeatures/issues/254

## Benchmark Results

- N/A -- new feature.


geom/dcel_extract.go
```diff
@@ -0,0 +1,295 @@
...
+	}
+}
+
+// TODO: Line extraction isn't working too well at the moment. It's currently
+// extracting each line individually, which isn't intended. It might be better
+// to return a []line here, and then construct back into LineString and
+// MultiLineString as a separate logical step since it seems tricky to do
+// inline.
+
+func (d *doublyConnectedEdgeList) extractLineStrings(include func(uint8) bool) ([]LineString, error) {
```
peterstace, line 204: Given that the extraction algorithm here doesn't work as intended, we can replace it with something simpler.

peterstace: Fixed.

Issue #56, opened 2019-09-05
Overhaul error system for easier programatic interogation


Den3:
There is a scenario that we would like to know the error specific to geometry, for example invalid geometry. However, there is no way to identify whether or not the errors come from geometry specific problem. for instance, with the struct below. I am unable to tell the errors from `json.Unmarshal()` is geom specific.
```go
type Place struct {
    name string `json:"name"`
    geometry geom.AnyGeometry `json:"geometry"`
}
```

Due to this reason, we have to work around this by separate structs to
```go
type Place struct{
    name string `json:"name"`
    geometry geom.AnyGeometry `json:"-"`
}

type ValidGeom struct {
    geometry geom.AnyGeometry `json:"geometry"` 
}
var place Place
if err := json.Unmarshal(data, &place); err != nil {
    return fmt.Errorf("name is invalid: %s", err)
}

var validGeom ValidGeom
if err := json.Unmarshal(data, &validGeom); err != nil {
    return fmt.Errorf("geom is invalid: %s", err)
}
```

peterstace:
In general, the error system needs an overhaul.

Right now, all of the errors returned are opaque, and would require string matching in order to programmatically handle various errors in different ways.

But really, there are a few different classes of errors:

- Parsing errors (WKTs and WKBs). E.g. when parsing `POINT((0.5 1.0))`.
- Validation errors for geometry construction, e.g. polygon has a non closed ring.
- Not Implemented style errors (for Intersection, Intersects, and Equals -- and maybe more going forward). The intention is that the usage of this would be eliminated once everything is implemented.
- Errors that get passed through from outside the library. E.g. when marshalling, the `io.Writer` could return an error.

This ticket can be for a wider error overhaul. The following Gophercon 2019 talk gives an interesting way to construct errors that can be easily interrogated, it may be worth designing something similar to that. https://www.youtube.com/watch?v=4WIhhzTTd0Y

peterstace:
Closed via #308

PR #307, opened 2020-12-09
Conventional WKT Error handling

peterstace:
## Description

- Changes WKT parsing to use conventional Go error handling.

- Previously, had used "monad" style error handling. While this was concise, it was error prone to get right and hard to follow the flow.

- This change is a precursor to using standardised error types everywhere (rather than using anonymous error values from `fmt.Errorf`).

## Check List

Have you:

- Added unit tests? N/A, uses existing.

## Related Issue

- Relates to https://github.com/peterstace/simplefeatures/issues/56

## Benchmark Results

- N/A, no benchmarks yet for WKT parsing.

### DIFF
--- a/geom/wkt_lexer.go
+++ b/geom/wkt_lexer.go
@@ -13,6 +13,7 @@ type wktLexer struct {
 func newWKTLexer(wkt string) wktLexer {
 	var scn scanner.Scanner
 	scn.Init(strings.NewReader(wkt))
+	scn.Mode = scanner.ScanInts | scanner.ScanFloats | scanner.ScanIdents
 	return wktLexer{scn: scn}
 }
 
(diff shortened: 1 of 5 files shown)


geom/wkt_lexer.go
```diff
@@ -1,54 +1,50 @@
...
+	if w.peeked != "" {
+		tok := w.peeked
+		w.peeked = ""
+		return tok
 	}
 
-	var err error
 	w.scn.Error = func(_ *scanner.Scanner, msg string) {
-		err = errors.New(msg)
+		panic(msg)
```
peterstace, line 27: TODO: need to investigate more in what cases this could happen (and make sure that it can't).

peterstace: Not using panic here anymore.

PR #366, opened 2021-06-11
Add Fuzz Testing for WKT parsing

peterstace:
## Description

Adds Fuzz testing for WKT parsing.

This requires the experimental branch `dev.fuzz` of the go toolchain to run (and is protected by a build tag so that it's not compiled using the mainline go tool).

The fuzz tests start with a corpus of WKT strings collected from the sourcecode of this repo. They then mutate the corpus (using increasing code coverage as a guiding force), checking to see if the `geom.UnmarshalWKT` function hits an infinite loop or panics.

## Check List

Have you:

- Added unit tests? N/A

## Related Issue

- N/A

## Benchmark Results

N/A

### DIFF
--- a/geom/wkt_fuzz_test.go
+++ b/geom/wkt_fuzz_test.go
@@ -1,0 +1,56 @@
+// +build gofuzzbeta
+
+package geom_test
+
+import (
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/peterstace/simplefeatures/geom"
+	"github.com/peterstace/simplefeatures/internal/extract"
+)
+
+func FuzzParseUnmarshalWKT(f *testing.F) {
+	corpus, err := extract.StringsFromSource("..")
+	if err != nil {
+		f.Fatalf("could not build corpus: %v", err)
+	}
+	for _, str := range corpus {
+		if allowInCorpus(str) {
+			f.Add(str)
+			f.Log(str)
+		}
+	}
+
+	f.Fuzz(func(t *testing.T, wkt string) {
+		done := make(chan struct{})
+		go func() {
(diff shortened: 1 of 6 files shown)


geom/wkt_fuzz_test.go
```diff
@@ -0,0 +1,56 @@
...
+import (
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/peterstace/simplefeatures/geom"
+	"github.com/peterstace/simplefeatures/internal/extract"
+)
+
+func FuzzParseUnmarshalWKT(f *testing.F) {
```
albertteoh, line 14: where is `testing.F` defined? I noticed there's a custom `gofuzzbeta` build tag here, maybe it's related somehow.

albertteoh (commented):
just checking, is this PR something you'd like reviewed or is it somewhat of a draft/WIP? If the former, I'll take a closer look.

peterstace:
> just checking, is this PR something you'd like reviewed or is it somewhat of a draft/WIP? If the former, I'll take a closer look.

This is a bit of a WIP/draft (I'll change the PR to draft status to make that explicit). I was basically playing around with the upcoming Golang [fuzzing](https://blog.golang.org/fuzz-beta) feature, so see if it would be useful for simplefeatures.

My gut feel is that it's probably _not_ terribly useful, although I think I need to come back and revisit in a bit more detail (hence I haven't closed the PR yet).

geom/wkt_fuzz_test.go
peterstace: Yes, that's right -- you need to use a custom branch of the Go toolchain to get this working (instructions [here](https://blog.golang.org/fuzz-beta)).

Issue #378, opened 2021-07-17
Add DumpPoints


peterstace:
https://postgis.net/docs/ST_DumpPoints.html

It should be implemented on every geometry type _except_ for `Point` (since it would be redundant).

It should preserve M and Z coordinates.

It shouldn't attempt to do any deduplication.

PR #398, opened 2021-08-04
Add `DumpCoordinates` methods

peterstace:
## Description

Adds `DumpCoordinates` methods on geometry types. These methods give back a
`Sequence` that contains all points in the geometry.

## Check List

Have you:

- Added unit tests? Yes.

## Related Issue

- Part of https://github.com/peterstace/simplefeatures/issues/378

## Benchmark Results

N/A

### DIFF
--- a/geom/type_coordinates.go
+++ b/geom/type_coordinates.go
@@ -1,6 +1,8 @@
 package geom
 
-// Coordinates represents a point location.
+// Coordinates represents a point location. Coordinates values may be
+// constructed manually using the type definition directly. Alternatively, one
+// of the New(XYZM)Coordinates constructor functions can be used.
 type Coordinates struct {
 	// XY represents the XY position of the point location.
 	XY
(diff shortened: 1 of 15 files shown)


geom/type_coordinates.go
```diff
@@ -18,3 +25,76 @@ type Coordinates struct {
 	// or not Z and M are populated.
 	Type CoordinatesType
 }
+
+// String gives a string representation of the coordinates.
+func (c Coordinates) String() string {
```
peterstace, line 30: Need to add a test for this.

geom/type_coordinates.go
```diff
@@ -18,3 +25,76 @@ type Coordinates struct {
...
+		return append(dst, c.X, c.Y, c.M)
+	case DimXYZM:
+		return append(dst, c.X, c.Y, c.Z, c.M)
+	default:
+		panic(c.Type.String())
+	}
+}
+
+// NewXYCoordinates constructs a new set of coordinates of type XY.
+func NewXYCoordinates(x, y float64) Coordinates {
```
peterstace, line 67: These probably aren't necessary anymore... I used then for tests before I switched from returning `[]Coordinates` to returning `Sequence`. Can be removed.

geom/type_geometry.go
```diff
@@ -767,3 +767,26 @@ func (g Geometry) appendDump(gs []Geometry) []Geometry {
...
+// DumpCoordinates returns the control points making up the geometry as a
+// Sequence.
+func (g Geometry) DumpCoordinates() Sequence {
+	switch g.gtype {
+	case TypeGeometryCollection:
+		return g.AsGeometryCollection().DumpCoordinates()
+	case TypePoint:
+		return g.AsPoint().DumpCoordinates()
+	case TypeLineString:
+		return g.AsLineString().Coordinates()
```
peterstace, line 780: There is no `DumpCoordinates` method for `LineString`, because the existing `Coordinates` method does the exact same thing. Even though it's a bit inconsistent, I think that's less confusing than having two methods that do exactly the same thing.

geom/type_coordinates.go
peterstace: Removed in 4f6f5b6

geom/type_coordinates.go
peterstace: Added in a986bf8

geom/type_geometry.go
albertteoh: if each geometry implemented `DumpCoordinates()`, could we do something like `g.AsDumpable().DumpCoordinates()` (for want of a better function name) and not required the switch-case block?

peterstace: That (or something similar) might be possible. The challenge would be implementing `AsDumpable()` on the `Geometry` type. The obvious way to implement it would be to use a `switch`, but that would defeat the point since we're trying to eliminate a switch statement. It _may_ be possible to use some sort of reflect/unsafe magic though.

The "switch" pattern actually shows up extremely frequently in this file (and is a bit of a code smell). Possibly it could be implemented generically so that we only need a single switch statement for all interface types.

Thanks for bringing this up! 😄 I've raised a ticket here: https://github.com/peterstace/simplefeatures/issues/399

Issue #396, opened 2021-07-27
Look for any places where NaN could leak into geometries


peterstace:
Consider Infs as well.

PR #401, opened 2021-08-07
Alter Internal Structure Of MultiPoint

peterstace:
## Description

- Alters the internals of `MultiPoint` to keep a slice of `Point` rather than a `Sequency` and a `BitSet` indicating empty points.

- This gives `MultiPoint` consistency with the other `Multi*` geometries (`MultiLineString` and `MultiPolygon`).

- It's also going to make implementing NaN and Inf checks easier later (see the linked ticket).

## Check List

Have you:

- Add cmprefimpl tests? (if appropriate?) N/A, relies on existing.

## Related Issue

- Part of https://github.com/peterstace/simplefeatures/issues/396

## Benchmark Results

  `run_benchmarks.sh` script.

TODO

<details>

```
PASTE BENCHMARKS HERE
```


geom/type_multi_point.go
```diff
@@ -9,14 +9,13 @@ import (
 // zero value is the empty MultiPoint (i.e. a collection of zero points) with
 // 2D coordinates type. It is immutable after creation.
 type MultiPoint struct {
-	seq   Sequence
-	empty BitSet
+	points []Point
+	ctype  CoordinatesType
```
peterstace, line 13: This is the start of the change -- everything else flows from there.

albertteoh:
CI is reporting this error: `cmprefimpl_1  | internal/cmprefimpl/cmpgeos/handle.go:296:10: undefined: geom.NewMultiPointFromPoints`

peterstace:
> CI is reporting this error: `cmprefimpl_1 | internal/cmprefimpl/cmpgeos/handle.go:296:10: undefined: geom.NewMultiPointFromPoints`

Ahh, you're right, thanks. Fixed in 2d01f13e025d9019c1d404719de2c8c54b6096e5

There's a breaking change there, I'll make sure to include this in the release notes.

geom/type_multi_point.go
```diff
@@ -9,14 +9,13 @@ import (
 // zero value is the empty MultiPoint (i.e. a collection of zero points) with
 // 2D coordinates type. It is immutable after creation.
 type MultiPoint struct {
-	seq   Sequence
-	empty BitSet
```
peterstace, line 13: `BitSet` is no longer used by anything, so I'll remove it from the codebase. Even though that increase the surface area of the API that is involved in the breaking change, I think that's better than leaving unused code around.

peterstace: Removed in a22a463d176504a3ed604b912d7453331fa6be4d

geom/type_multi_point.go
```diff
@@ -26,39 +25,11 @@ func NewMultiPointFromPoints(pts []Point, opts ...ConstructorOption) MultiPoint
...
-// point.
-func NewMultiPointWithEmptyMask(seq Sequence, empty BitSet, opts ...ConstructorOption) MultiPoint {
-	return MultiPoint{
-		seq,
-		empty.Clone(), // clone so that the caller doesn't have access to the internal empty set
+	cp := make([]Point, len(pts))
+	for i := range cp {
+		cp[i] = pts[i].ForceCoordinatesType(ctype)
 	}
+	return MultiPoint{cp, ctype}
```
albertteoh, line 32: this block of code is quite similar to `ForceCoordinatesType`; perhaps there's an opportunity to reduce duplication?

albertteoh (approved):
lgtm!

geom/type_multi_point.go
peterstace: That's a good catch.

The most obvious way to reduce the duplication is to just do `return MultiPoint{pts, ctype}.ForceCoordinatesType(ctype)`, but that has the awkward side effect that some of the invariant of the temporary `MultiPoint` don't hold (i.e. the `ctype` doesn't necessarily match that of each `Point` in the slice). Even though that invariant is only broken very temporarily, it makes me a bit uneasy.

So instead, in 9cb5049ff00ac71a41243317fa1ee0b062cebec9 I've split out the "force a slice of Points to have a particular ctype" logic into a new function, and am calling that from the `NewMultiPoint` function and the `ForceCoordinatesType` method.

peterstace: As an aside, the invariant wasn't explicitly documented anywhere. I've added documentation in 0b5e8bd247fbb9c0f441fa90fd18916555a8e731

geom/geojson_unmarshal.go
```diff
@@ -263,8 +263,19 @@ func geojsonNodeToGeometry(node interface{}, ctype CoordinatesType, opts []Const
...
+			return MultiPoint{}.ForceCoordinatesType(ctype).AsGeometry(), nil
+		}
+		points := make([]Point, len(node.coords))
+		for i, coords := range node.coords {
+			coords, ok := oneDimFloat64sToCoordinates(coords, ctype)
+			if ok {
+				points[i] = NewPoint(coords, opts...)
+			} else {
+				points[i] = NewEmptyPoint(ctype)
+			}
```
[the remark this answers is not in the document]
peterstace: That's a pretty good idea. We could have something like:

```
// OptionalCoordinates represents Coordinates that may optionally be empty.
type OptionalCoordinates struct {
    // Coordinates hold the full coordinates when they are non-empty, or just
    // the CoordinateType when they are empty.
    Coordinates

    // Empty indicates if the XY, Z, and M parts of the Coordinates are empty
    // or populated.
    Empty bool
}
```

We could use this new type for the internals of `Point` as well. Since it will collapse together `NewPoint` and `NewEmptyPoint` (and there may be some fairly noisy flow on effects), I'll raise a new ticket for this and work on it before the NaN/Infinity checks.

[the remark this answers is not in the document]
peterstace: I've created a new ticket here: https://github.com/peterstace/simplefeatures/issues/402

peterstace:
Thanks for the review!

albertteoh (approved):
thanks, lgtm!

Issue #402, opened 2021-08-09
Collapse Point constructors


peterstace:
See https://github.com/peterstace/simplefeatures/pull/401#discussion_r685096260 for more context.

There are two main changes we want:

- Create a new `OptionalCoordinates` type:

```
// OptionalCoordinates represents Coordinates that may optionally be empty.
type OptionalCoordinates struct {
    // Coordinates hold the full coordinates when they are non-empty, or just
    // the CoordinateType when they are empty.
    Coordinates

    // Empty indicates if the XY, Z, and M parts of the Coordinates are empty
    // or populated.
    Empty bool
}
```

- Collapse together the `NewPoint`, `NewEmptyPoint`, and `NewPointFromXY` functions into one. `NewPointFromXY` could have been collapsed into `NewPoint` without adding `OptionalCoordinates`, but that's something that would be nice to do at the same time since it's also a breaking change. All other geometry types just have a single constructor, so it was a  bit weird that Point had multiple.

- Clean up any flow on from collapsing the constructors. This includes GeoJSON parsing, WKT parsing, and WKB parsing. We should be able to remove a bunch of special "empty" cases for Points and MultiPoints.

PR #404, opened 2021-08-11
Collapse Point Constructors

peterstace:
## Description

- Collapses the 3 point constructor functions (`NewEmptyPoint`, `NewPoint`, and
  `NewPointFromXY`) into a single constructor function.

- Introduce a new `OptionalCoordinates` type, which represents coordinates that
  may or may not be empty. This is used in the new point constructor.

## Check List

Have you:

- Added unit tests? Relies on existing (and updated) unit tests.

## Related Issue

Fixes https://github.com/peterstace/simplefeatures/issues/402

## Benchmark Results

N/A

### DIFF
--- a/geom/type_point.go
+++ b/geom/type_point.go
@@ -19,18 +19,9 @@ type Point struct {
 }
 
 // NewPoint creates a new point gives its Coordinates.
-func NewPoint(c Coordinates, _ ...ConstructorOption) Point {
-	return Point{c, true}
-}
-
-// NewEmptyPoint creates a Point that is empty.
-func NewEmptyPoint(ctype CoordinatesType) Point {
-	return Point{Coordinates{Type: ctype}, false}
-}
-
-// NewPointFromXY creates a new point from an XY.
-func NewPointFromXY(xy XY, _ ...ConstructorOption) Point {
-	return Point{Coordinates{XY: xy, Type: DimXY}, true}
+func NewPoint(oc OptionalCoordinates, _ ...ConstructorOption) Point {
+	c := Coordinates{oc.XY, oc.Z, oc.M, oc.Type}
+	return Point{c, !oc.Empty}
 }
 
 // Type returns the GeometryType for a Point
@@ -168,7 +159,8 @@ func (p Point) TransformXY(fn func(XY) XY, opts ...ConstructorOption) (Point, er
 	}
 	newC := p.coords
 	newC.XY = fn(newC.XY)
-	return NewPoint(newC, opts...), nil
+	newOC := newC.AsOptionalCoordinates()
+	return NewPoint(newOC, opts...), nil
 }
 
 // Centroid of a point is that point.
@@ -196,7 +188,7 @@ func (p Point) CoordinatesType() CoordinatesType {
 // is added, then new values are populated with 0.
 func (p Point) ForceCoordinatesType(newCType CoordinatesType) Point {
 	if !p.full {
-		return NewEmptyPoint(newCType)
+		return Point{coords: Coordinates{Type: newCType}}
 	}
 	if newCType.Is3D() != p.coords.Type.Is3D() {
 		p.coords.Z = 0
--- a/README.md
+++ b/README.md
@@ -177,7 +177,8 @@ Encoding and decoding WKB directly:
 
 ```go
 // Marshal as WKB
-pt := geom.NewPointFromXY(geom.XY{1.5, 2.5})
+coords := geom.OptionalCoordinates{XY: geom.XY{1.5, 2.5}}
+pt := geom.NewPoint(coords)
 wkb := pt.AsBinary()
 fmt.Println(wkb) // Prints: [1 1 0 0 0 0 0 0 0 0 0 248 63 0 0 0 0 0 0 4 64]
 
@@ -199,7 +200,8 @@ db.Exec(`
 )
 
 // Insert our geometry and population data into PostGIS via WKB.
-nyc := geom.NewPointFromXY(geom.XY{-74.0, 40.7})
+coords := geom.OptionalCoordinates{XY: geom.XY{-74.0, 40.7}}
+nyc := geom.NewPoint(coords)
 db.Exec(`
     INSERT INTO my_table
     (my_geom, population)
--- a/geom/accessor_test.go
+++ b/geom/accessor_test.go
@@ -25,10 +25,10 @@ func TestLineStringAccessor(t *testing.T) {
 	pt56 := xyCoords(5, 6)
 
 	t.Run("start", func(t *testing.T) {
-		expectGeomEq(t, ls.StartPoint().AsGeometry(), NewPoint(pt12).AsGeometry())
+		expectGeomEq(t, ls.StartPoint().AsGeometry(), NewPoint(pt12.AsOptionalCoordinates()).AsGeometry())
 	})
 	t.Run("end", func(t *testing.T) {
-		expectGeomEq(t, ls.EndPoint().AsGeometry(), NewPoint(pt56).AsGeometry())
+		expectGeomEq(t, ls.EndPoint().AsGeometry(), NewPoint(pt56.AsOptionalCoordinates()).AsGeometry())
 	})
 	t.Run("num points", func(t *testing.T) {
 		expectIntEq(t, seq.Length(), 3)
--- a/geom/alg_convex_hull.go
+++ b/geom/alg_convex_hull.go
@@ -17,7 +17,7 @@ func convexHull(g Geometry) Geometry {
 
 	// Check for point case:
 	if !hasAtLeast2DistinctPointsInXYs(pts) {
-		return NewPointFromXY(pts[0]).AsGeometry()
+		return pts[0].AsPoint().AsGeometry()
 	}
 
 	hull := monotoneChain(pts)
--- a/geom/alg_intersection.go
+++ b/geom/alg_intersection.go
@@ -16,9 +16,9 @@ func intersectionOfIndexedLines(
 				return nil
 			}
 			if inter.ptA == inter.ptB {
-				if pt := inter.ptA; !seen[pt] {
-					pts = append(pts, NewPointFromXY(pt))
-					seen[pt] = true
+				if xy := inter.ptA; !seen[xy] {
+					pts = append(pts, xy.AsPoint())
+					seen[xy] = true
 				}
 			} else {
 				lss = append(lss, line{inter.ptA, inter.ptB}.asLineString())
@@ -41,7 +41,7 @@ func intersectionOfMultiPointAndMultiPoint(mp1, mp2 MultiPoint) MultiPoint {
 	for i := 0; i < mp2.NumPoints(); i++ {
 		xy, ok := mp2.PointN(i).XY()
 		if ok && inMP1[xy] {
-			pts = append(pts, NewPointFromXY(xy))
+			pts = append(pts, xy.AsPoint())
 		}
 	}
 	return NewMultiPoint(pts)
--- a/geom/alg_point_on_surface.go
+++ b/geom/alg_point_on_surface.go
@@ -127,7 +127,7 @@ func pointOnAreaSurface(poly Polygon) (Point, float64) {
 	}
 	midX := (bestA + bestB) / 2
 
-	return NewPointFromXY(XY{midX, midY}), bestB - bestA
+	return XY{midX, midY}.AsPoint(), bestB - bestA
 }
 
 func sortAndUniquifyFloats(fs []float64) []float64 {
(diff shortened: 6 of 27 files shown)


geom/type_point.go
```diff
@@ -19,18 +19,9 @@ type Point struct {
...
-func NewEmptyPoint(ctype CoordinatesType) Point {
-	return Point{Coordinates{Type: ctype}, false}
-}
-
-// NewPointFromXY creates a new point from an XY.
-func NewPointFromXY(xy XY, _ ...ConstructorOption) Point {
-	return Point{Coordinates{XY: xy, Type: DimXY}, true}
+func NewPoint(oc OptionalCoordinates, _ ...ConstructorOption) Point {
+	c := Coordinates{oc.XY, oc.Z, oc.M, oc.Type}
+	return Point{c, !oc.Empty}
```
peterstace, line 24: Something to point out here is that the `bool` used to indicate empty in `Point` and in `OptionalCoordinates` has flipped meaning. In `Point`, the bool is `full` and in `OptionalCoordinates` the bool is `Empty`.

There are a few reasons for this:

- It's important for the zero value of `Point` to represent `POINT EMPTY`. The zero value for all other concrete geometry types is the empty XY value of that geometry. So therefore `false` must encode emptiness.
- When specifying an `OptionalCoordinates`, it's intended that users can specify _either_ `Empty: true` _or_ `XY: geom.XY{X:2, Y:3}` (and then also `Z` and `M` if appropriate). This makes things a bit more ergonomic from the perspective of users. If the `bool` had the same meaning as in `Point`, then the user would have to specify `Populated: true, XY: geom.XY{...}` for populated points, and no fields at all for empty points. It could be fairly easy for users to miss the `Populated: true` part, and _think_ they are creating a populated point (but really it would be empty).

geom/type_point.go
```diff
@@ -19,18 +19,9 @@ type Point struct {
...
-
-// NewEmptyPoint creates a Point that is empty.
-func NewEmptyPoint(ctype CoordinatesType) Point {
-	return Point{Coordinates{Type: ctype}, false}
-}
-
-// NewPointFromXY creates a new point from an XY.
-func NewPointFromXY(xy XY, _ ...ConstructorOption) Point {
-	return Point{Coordinates{XY: xy, Type: DimXY}, true}
+func NewPoint(oc OptionalCoordinates, _ ...ConstructorOption) Point {
```
albertteoh, line 22: could `func NewPoint(oc *Coordinates, ...)` be an option? What are the drawbacks of this option in your opinion?

peterstace: We'd still need to specify the `CoordinatesType`, since `POINT EMPTY` is distinct from other types of empty points, such as `POINT Z EMPTY`. That's probably not too much of a barrier though, since you could make it `NewPoint(ctype CoordinatesType, oc *Coordinates, ...)` (and remove the coordinates type from `Coordinates`).

So something like that _could_ be workable from a technical standpoint. Indeed, a lot of codebases use pointers to indicate optionality, so there's good precedence for it.

I'm not a huge fan of encoding optional using pointers, because IMO it leads to confusing library APIs. In Go (probably not limited to just Go), pointers can semantically mean a few different things:

- Immutable sharing i.e. the each part of the code that has a copy of the pointer is free to read the value being pointed to. This is usually done for performance reasons.
- Mutable sharing i.e. each part of the code having a copy of the pointer can write to the value pointed to (with synchronization if appropriate). This seems most common in the internals of data structures, or for business/product style code where multiple actors need to change the same state. Method receivers that are pointers also would fall into this category.
- Optionality i.e. _value semantics_ are desired, but the value is optional. In languages with more complex type systems, pointers wouldn't be used for this. E.g. C++ would use `std::optional<Coordinates>` and Haskell would use `Maybe Coordinates`.

IMO, code with pointers is already confusing enough, so I prefer to avoid it for optionality in cases where there are good alternatives. There are also performance impacts of using pointers when sharing semantics aren't required. They usually (but not always) cause the pointed to value to end up on the heap, which has GC implications. That's fine if the number of objects is limited, but for something like a `Point` where a program could have a many of them, it might cause some performance problems (I don't want to overstate the performance aspect -- my main concern is about making a confusing library API).

Alternate approaches to encode optionality:

- Rely on the zero value to indicate that the value hasn't been set. It doesn't work in this case, since the zero value is actually valid.
- Use extra bools to indicate if the value is present or not. This works great for the internals of types (e.g. `full` _inside_ `Point`) because you can hide the details and it's all pretty well encapsulated. It can be a tiny bit awkward for interfaces though, which I think is probably where you're coming from?
- Use separate methods for each combination of the options. This is ok if the number of options is limited, and is similar to what was used before this change (i.e. separate `NewPoint` and `NewEmptyPoint`).

peterstace: With all of that being said, I'm not 100% convinced that the functionality in this PR is actually making things that much better. The usage of `OptionalCoordinates` seems like a lot to wrap your head around as a user, and we have to rely on comments in that type explain what should and should not be populated in various empty vs non-empty scenarios. Having to explain that sort of stuff in comments in a bit of a flag that things might be tricky for users.

I still think that it's a good idea to collapse the `NewPoint` and `NewPointFromXY` constructors together (they're fairly redundant). But maybe it makes sense to retain `NewEmptyPoint`? The advantage of doing that would be:

- That's a nicer way to deal with optionality. `NewPoint` will accept a regular `Coordinates` value and `NewEmptyPoint` would accept a `CoordinatesType`.
- When we get to NaN/Inf checking (which is what I plan to take a look at next), we'll need to return an error from `NewPoint`. In the case of an empty `Point`, there could never actually be an error. So by having a different constructor for empties, the user won't have to handle errors in that case.

The downside is that users of `Point` (including the WKT/WKB/GeoJSON parsing code in this library) will need to use if statements to differentiate the empty/non-empty cases, so that they can call the appropriate constructor. Maybe that's a small price to pay though, especially since it's inside a library. I don't anticipate that users would _typically_ need to construct a value from raw numbers -- most would be decoding from another format. I'm feeling a bit torn, thoughts?

albertteoh: Thanks for the thoughts @peterstace, I agree that a `nil` parameter isn't a good option. It requires users to understand the meaning of passing `nil` and I think would be annoying/confusing to do this each time they want an "empty" `Pointer`.

> It can be a tiny bit awkward for interfaces though, which I think is probably where you're coming from?

I'm not sure, could you give an example where this would be awkward for interfaces? 

What I was thinking before was that it would be ideal if all geoms implemented a common interface (e.g. exposing funcs like `Summary()`, `String()`, etc.) which _should_ avoid the need for the `switch` blocks.

> With all of that being said, I'm not 100% convinced that the functionality in this PR is actually making things that much better. The usage of OptionalCoordinates seems like a lot to wrap your head around as a user, and we have to rely on comments in that type explain what should and should not be populated in various empty vs non-empty scenarios.

+1. I completely agree. IMO this concept/knowledge _should_ ideally be abstracted away from the user. i.e "As a user, I just want a `Point`. If I know the coordinates of it now, I'll give it to you, otherwise, just create one without coords for me". Our library should deal with the nuances of empty vs. non-empty points behind the scenes. Please correct me if I'm wrong here.

... and yes, perhaps just having `NewEmptyPoint` and `NewPoint` could be the best option.

> I don't anticipate that users would typically need to construct a value from raw numbers -- most would be decoding from another format. I'm feeling a bit torn, thoughts?

If most users do not construct a `Point` from raw numbers, would removing the `New...Point` ctors be a valid alternative? I'm assuming you meant most use cases involve creating a new `Point` via the WKT.

---

I'd like to list out the requirements for the `Point` constructor first to make sure I understand it correctly:
- There are two possible states of a Point: `empty` and `full`. 
- It must communicate to the user that the `CoordinateType` is a mandatory parameter that needs to be set explicitly. Making this optional makes it unclear what the `CoordinateType` will be if unset.
- It must communicate to the user that `Coordinates` are optional.
- `0` is a valid value for the X/Y-coordinate (and Z/M values), and so cannot rely on not setting `Coordinates` as a means of creating an "empty" `Point`.
- Be able to convey errors on NaN/Inf.

Our design goal is to have a clear and easy means of creating a `Point` that satisfies the above requirements.

It seems this use case is similar to that described in Dave Cheney's [functional options blog post](https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis), and it could be another alternative that we could consider.

For example, very hypothetically, if we were to ignore the (currently unused) `_ ...ConstructorOption`), we could do something like:
```
// NewPoint creates a new point gives its Coordinates.
func NewPoint(ctype CoordinatesType, opts ...func(*Point)) (*Point, error) {
	p := &Point{coords: Coordinates{Type: ctype}}
	for _, opt := range opts {
		if err := opt(p); err != nil {
                     return nil, err
                 }
	}
	return p, nil
}

func WithCoordinates(c *Coordinates) func(p *Point) error {
	return func(p *Point) error {
                if c == nil {
                       return nil
                }
                // check for NaN/Inf and return error if so...
		p.coords = c
		p.full = true
                 return nil
	}
}
```

So we create an "empty" `Point` with:
```
emptyPoint, err := NewPoint(newCType)
```

... and a "full" `Point` with:
```
fullPoint, err := NewPoint(DimXY, WithCoordinates(&Coordinates{Type: DimXY, XY: XY{1, 2}}))
```

A "setter" could also be used, but I feel it doesn't clearly communicate the configurability of `Coordinates` at construction.

So to address the original comment: https://github.com/peterstace/simplefeatures/pull/401#discussion_r685096260 we could do something like:
```
coords := oneDimFloat64sToCoordinates(coords, ctype)
// return nil coords here if no coords found.
points[i], err := NewPoint(coords, WithCoordinates(coords))
```

For the above, I think `nil` is a more valid means of conveying optionality; i.e. it should _never_ be the case that users are expected to supply `nil`, but more for internal lib use.

One apparent downside of this approach is that the `NewPointer` ctor already takes in variadiac args; although I see they're currently not used in the ctor.

What do you think?

albertteoh: > IMO this concept/knowledge should ideally be abstracted away from the user.

I just realised that this statement is probably incorrect given `"POINT EMPTY"` is a legitimate way for users to create a `Point`, and so I think #405 is the right approach.

peterstace: > > It can be a tiny bit awkward for interfaces though, which I think is probably where you're coming from?
>
> I'm not sure, could you give an example where this would be awkward for interfaces? 

Sorry, I wasn't really too clear in my comment. By "interfaces", I'm more
broadly referring to library API interfaces rather than Go interfaces. So the
awkwardness I'm referring to is for users to have to specify a `bool` member
within a field to indicate whether they have populated _another_ field or not.

> Our library should deal with the nuances of empty vs. non-empty points behind the scenes. Please correct me if I'm wrong here.

You're correct, especially when the user is creating a geometry from a WKT,
WKB, or GeoJSON. The only exception is when the user is constructing Points (or
other geometries) directly from `float64`s. In that case, they will have to be
concious of it (since empty points require 0 `float64`s and non-empty points
require 2, 3, or 4 `float64`s depending on the coordinates type).

> > I don't anticipate that users would typically need to construct a value from raw numbers -- most would be decoding from another format.

> If most users do not construct a `Point` from raw numbers, would removing the `New...Point` ctors be a valid alternative? I'm assuming you meant most use cases involve creating a new `Point` via the WKT.

Yes, _most_ cases would involve creating a `Point` from WKT, WKB, or GeoJSON.
We'll still need some sort of manual constructor though, since users will still
need to create points manually from `float64`s sometimes (it's just that it
would be a fairly rare usecase).

> I'd like to list out the requirements for the `Point` constructor first to make sure I understand it correctly:
> - There are two possible states of a Point: `empty` and `full`. 
> - It must communicate to the user that the `CoordinateType` is a mandatory parameter that needs to be set explicitly. Making this optional makes it unclear what the `CoordinateType` will be if unset.
> - It must communicate to the user that `Coordinates` are optional.
> - `0` is a valid value for the X/Y-coordinate (and Z/M values), and so cannot rely on not setting `Coordinates` as a means of creating an "empty" `Point`.
> - Be able to convey errors on NaN/Inf.
> 
> Our design goal is to have a clear and easy means of creating a `Point` that satisfies the above requirements.

Yes, that's correct.

> It seems this use case is similar to that described in Dave Cheney's [functional options blog post](https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis), and it could be another alternative that we could consider.

Yes, functional options would in general satisfy the constructor requirements
you outlined.

> A "setter" could also be used, but I feel it doesn't clearly communicate the configurability of `Coordinates` at construction.

Yes, the lack of clarify in communication is one downside to that. The other
downside is that _all_ geometry types so far are immutable after creation. So
adding a setter would make `Point` operate differently from the other geometry
types.

> One apparent downside of this approach is that the `NewPointer` ctor already takes in variadiac args; although I see they're currently not used in the ctor.

I think that's probably the biggest downside. While it's not used _right now_,
we'll have to make use of it when doing NaN and Inf checks. Well either want to
skip those validations when the user supplies the `DisableAllValidations`
option.

peterstace:
Closing this out, since this was solved in https://github.com/peterstace/simplefeatures/pull/405 instead.

PR #405, opened 2021-08-13
Remove `NewPointFromXY` constructor

peterstace:
## Description

This constructor has too much redundancy with the `NewPoint` constructor.  The `NewPoint` constructor is now supplemented by a new `AsPoint` helper method on the `XY` type.

Note that the method signature of `AsPoint` and `NewPoint` will need to change to additionally return an error as part of https://github.com/peterstace/simplefeatures/issues/396 (so that will need to get done before the next tag).

## Check List

Have you:

## Related Issue

Part of https://github.com/peterstace/simplefeatures/issues/402, a less extreme alternative to the approach taken in PR https://github.com/peterstace/simplefeatures/pull/404

There is some discussion about this approach in this comment thread: https://github.com/peterstace/simplefeatures/pull/404#discussion_r687585264

## Benchmark Results

N/A

### DIFF
--- a/README.md
+++ b/README.md
@@ -177,7 +177,8 @@ Encoding and decoding WKB directly:
 
 ```go
 // Marshal as WKB
-pt := geom.NewPointFromXY(geom.XY{1.5, 2.5})
+coords := geom.Coordinates{XY: geom.XY{1.5, 2.5}}
+pt := geom.NewPoint(coords)
 wkb := pt.AsBinary()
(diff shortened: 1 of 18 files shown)


albertteoh (approved):
LGTM, I like that it removes another constructor by inverting the construction of a `Point` via the `Coordinate`.

peterstace:
Thanks for the review!

PR #407, opened 2021-08-16
Use TB Interface For Test Helpers

peterstace:
## Description

These test helpers are useful for more than just unit tests. E.g. they can be used to verify that preliminary benchmark setup is correct. I needed this in a separate PR, but added it here instead since it's a fairly separate change.

## Check List

Have you:

- Added unit tests? N/A

## Related Issue

- N/A

## Benchmark Results

N/A

### DIFF
--- a/geom/util_test.go
+++ b/geom/util_test.go
@@ -8,7 +8,7 @@ import (
 	. "github.com/peterstace/simplefeatures/geom"
 )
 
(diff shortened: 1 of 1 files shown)


albertteoh (approved):

peterstace:
Thanks for the review!

PR #409, opened 2021-08-21
Add golangci lint

### LINKED ISSUE #408: Add golangci-lint to CI
`golangci-lint` shows a few problems.

Steps:

- Fix any bugs or major problems discovered by `golangci-lint` (from skimming the list, there might be one or two bugs).
- Add `golangci-lint` to CI.

albertteoh:
## Description

Adds some basic golangci-lint checks to the CI build, and fixes any existing linter issues discovered.

## Check List

Have you:

~- Added unit tests? ~

~- Add cmprefimpl tests? (if appropriate?)~

## Related Issue

- Fixes #408.


geom/alg_distance.go
```diff
@@ -35,7 +35,6 @@ func Distance(g1, g2 Geometry) (float64, bool) {
 	if len(xys1)+len(lns1) > len(xys2)+len(lns2) {
 		xys1, xys2 = xys2, xys1
 		lns1, lns2 = lns2, lns1
-		g1, g2 = g2, g1
```
albertteoh, line 38: `g1` and `g2` are never used after this line.

geom/dcel_label.go
```diff
@@ -61,6 +61,7 @@ func newLocationsOnBoundary(operand operand) [2]location {
 	return locs
 }
 
+// nolint
```
albertteoh, line 64: quite a few constants in this block are not used.

geom/util_test.go
```diff
@@ -135,6 +135,7 @@ func expectTrue(t testing.TB, got bool) {
 	expectBoolEq(t, got, true)
 }
 
+// nolint: deadcode, unused
```
albertteoh, line 138: thought this func might still be useful for future test assertions.

geom/validation_test.go
```diff
@@ -8,10 +8,6 @@ import (
 	. "github.com/peterstace/simplefeatures/geom"
 )
 
-func xy(x, y float64) Coordinates {
```
albertteoh, line 11: I think your recent changes may have made this redundant? if you feel it's still needed, I can add it back in.

internal/cmprefimpl/cmppg/checks.go
```diff
@@ -502,65 +502,3 @@ func containsOnlyPolygonsOrMultiPolygons(g geom.Geometry) bool {
 		return false
 	}
 }
-
-func checkForceCoordinatesDimension(t *testing.T, want UnaryResult, g geom.Geometry) {
```
albertteoh, line 506: let me know if still needed.

rtree/nearest.go
```diff
@@ -6,11 +6,7 @@ import "container/heap"
 // as measured by the Euclidean metric. Note that there may be multiple records
 // that are equidistant from the input box, in which case one is chosen
 // arbitrarily. If the RTree is empty, then false is returned.
-func (t *RTree) Nearest(box Box) (int, bool) {
```
albertteoh, line 9: not caught by linter but noticed it could be simplified.

geom/dcel_label.go
peterstace: I think we can just remove the unused constants here, but happy to do that as a second pass in order to get the base golangci lint change out first.

internal/cmprefimpl/cmppg/checks.go
peterstace: Ahh, this actually is still needed. It was a bug that it wasn't being invoked by anything. Can you add this code back and call it from the list of checks [here](https://github.com/peterstace/simplefeatures/blob/master/internal/cmprefimpl/cmppg/fuzz_test.go#L60)?

rtree/nearest.go
peterstace: According to [CodeReviewComments](https://github.com/golang/go/wiki/CodeReviewComments#named-result-parameters), they reckon it's better _not_ to use named result parameters in order just to avoid declarations inside the function. I agree with that.

_However_, they also say it's ok to use named result parameters to add clarity to the meaning of the return parameters when it's not clear. It could be argued that it's not clear from the method signature than an `int` is a record ID.

This change is good due to the second reason, but (IMO) not the first reason. I'm happy to keep it in. Thoughts?

internal/cmprefimpl/cmppg/fuzz_test.go
```diff
@@ -40,7 +40,7 @@ func TestFuzz(t *testing.T) {
 				t.Skip("Causes unmarshalling to fail for derivative geometry")
 			}
 
-			want, err := BatchPostGIS{pg.db}.Unary(g)
+			want, err := BatchPostGIS(pg).Unary(g)
```
peterstace, line 43: Does this change compile? I haven't run it through the full CI, but when I manually make this change to my local I get: `cannot convert pg.db (type *sql.DB) to type BatchPostGIS`.

peterstace (changes requested):
Awesome, thanks for doing this!

I _think_ there's just 1 compile error that should be addressed (the DB handle conversion). I might be mistaken about that though.

.github/workflows/ci.yml
```diff
@@ -1,6 +1,7 @@
 name: build
 on:
-  push
+  push:
+  pull_request:
```
albertteoh, line 4: fixes the problem of PR builds not running.

.github/workflows/golangci-lint.yml
```diff
@@ -0,0 +1,17 @@
+name: golangci-lint
+on:
+  push:
+  pull_request:
```
albertteoh, line 4: this fixes the linting on forks/branches so devs won't need to create a PR in order for the builds and linters to run.

albertteoh:
I've address the review comments. Sorry for the multiple commits, I was testing out the fix to getting builds working from branches for PRs.

rtree/nearest.go
peterstace: All fair points, agreed 👍

internal/cmprefimpl/cmppg/fuzz_test.go
peterstace: You're right 👍

geom/dcel_label.go
```diff
@@ -61,6 +61,7 @@ func newLocationsOnBoundary(operand operand) [2]location {
 	return locs
 }
 
+//nolint
```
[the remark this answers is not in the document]
peterstace: Ahh, yes can you remove these ones?

We used to have a raw uint8 to track a bunch of flags at the same time, but it was a premature optimisation that didn't help much (and was very difficult to work with). We're now using explicit flags and enums in the `label` and `location` types a bit further up in this file.

peterstace (approved):
> Sorry for the multiple commits

All good 😃 

LGTM (although we can remove the unused consts in `dcel_label.go`)

geom/dcel_label.go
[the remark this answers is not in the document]
albertteoh: For sure, done in [1efc65e](https://github.com/peterstace/simplefeatures/pull/409/commits/1efc65ea0ef5111c66cc21dcccbe9c1247aec5ba).

albertteoh:
Thanks! 🙏

PR #413, opened 2021-08-30
Rename Geometry Constructor Methods For Consistency

peterstace:
## Description

The vast majority of constructors are now in the form `NewFoo`, where `Foo` is the name of the thing being constructed (e.g. a geometry type such as `MultiPolygon`).

Previously, the constructors were named so that the type that was being constructed _from_ was also mentioned.

E.g. `NewPolygonFromRings` was renamed to `NewPolygon`. The `FromRings` part really was redundant, since polygons are always constructed from their constituent rings (there really is no other practical way to construct them).

## Check List

Have you:

- Added unit tests? N/A, just a rename.

## Related Issue

N/A

## Benchmark Results

N/A

### DIFF
--- a/geom/alg_convex_hull.go
+++ b/geom/alg_convex_hull.go
@@ -40,7 +40,7 @@ func convexHull(g Geometry) Geometry {
 	if err != nil {
 		panic(fmt.Errorf("bug in monotoneChain routine - didn't produce a valid ring: %v", err))
 	}
-	poly, err := NewPolygonFromRings([]LineString{ring})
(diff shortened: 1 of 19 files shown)


albertteoh (approved):
I like it 👌

peterstace:
The build is failing due to a 504 from coveralls... The actual main CI part is working though. I'm going to ignore it for now, and investigate what's going wrong there if it keeps happening (I'm guessing it's _probably_ a transient problem).

peterstace:
Thanks for the speedy review @albertteoh 🥳

Issue #412, opened 2021-08-28
Consider whether or not Envelope should be able to be empty


peterstace:
There interface would be much nicer in many cases if the `Envelope` type was allowed to be empty, similar to regular geometries.

This is somewhat related to https://github.com/peterstace/simplefeatures/issues/410 (in fact, if we make Envelopes possibly empty, that would eliminate the need for 410).

PR #414, opened 2021-09-05
Allow Envelopes to be optionally empty

peterstace:
## Description

This is still a **WIP/DRAFT**.

Allow Envelopes to be optionally empty.

Envelopes now act more like regular geometries, in in the sense that they can be empty or not. This helps to eliminate a large amount of bool checking code outside of the envelope whenever envelopes are used.

## Check List

Have you:

- Added unit tests? Yes.

- Add cmprefimpl tests? (if appropriate?) No, relies on existing.

## Related Issue

- https://github.com/peterstace/simplefeatures/issues/412

## Benchmark Results

- [ ] **TODO**

  `run_benchmarks.sh` script.

<details>

```
PASTE BENCHMARKS HERE
```

</details>


peterstace:
Still need to write release notes:

- [ ] Add release notes.

CHANGELOG.md
```diff
@@ -1,5 +1,41 @@
...
+- **Breaking change**: The `Envelope` type can now be an empty envelope.
+  Previously, it was only able to represent a rectangle with some area, a
+  horizontal or vertical line, or a single point. Its `AsGeometry` returns
+  an empty `GeometryCollection` in the case where it's empty. The result of
+  `AsGeometry` is unchanged for non-empty envelopes.
+
+- **Breaking change**: The `NewEnvelope` function signature has changed. It now
+  accepts a slice of `geom.XY` as the sole argument. The behaviour of the
+  function is the same as before, except that if no XY values are provided then
+  an empty envelope is returned.
```
albertteoh, line 14:
```suggestion
  an empty envelope is returned without error.
```

geom/type_envelope.go
```diff
@@ -7,28 +7,30 @@ import (
...
-func NewEnvelope(first XY, others ...XY) (Envelope, error) {
-	if err := first.validate(); err != nil {
-		return Envelope{}, err
-	}
-	env := Envelope{
-		min: first,
-		max: first,
-	}
-	for _, xy := range others {
+func NewEnvelope(xys []XY) (Envelope, error) {
```
albertteoh, line 31: This seems more intuitive and simpler 👍🏼

geom/type_envelope.go
```diff
@@ -7,28 +7,30 @@ import (
...
+// An envelope can be thought of as as being similar to a regular geometry, but
+// can only represent an empty geometry, a single point, a horizontal or
+// vertical line, or an axis aligned rectangle with some area.
+//
+// The Envelope zero value is the empty envelope. Envelopes are immutable after
+// creation.
 type Envelope struct {
-	min XY
-	max XY
+	full bool
```
albertteoh, line 23: Curious, what were the reasons for choosing `full` instead of `empty`?

internal/cmprefimpl/cmpgeos/handle.go
```diff
@@ -469,94 +469,90 @@ func (h *Handle) dimension(g geom.Geometry) (int, error) {
 	return dim, nil
 }
 
-func (h *Handle) envelope(g geom.Geometry) (geom.Envelope, bool, error) {
+func (h *Handle) envelope(g geom.Geometry) (geom.Envelope, error) {
```
albertteoh, line 472: Nice 👍

geom/util_test.go
```diff
@@ -168,6 +174,25 @@ func expectBytesEq(t testing.TB, got, want []byte) {
...
+	t.Helper()
+	if got != want {
+		t.Errorf("\ngot:  %v\nwant: %v\n", got, want)
+	}
+}
+
+func expectEnvEq(t testing.TB, got, want Envelope) {
+	t.Helper()
+	if !ExactEquals(got.Min().AsGeometry(), want.Min().AsGeometry()) ||
+		!ExactEquals(got.Max().AsGeometry(), want.Max().AsGeometry()) {
```
albertteoh, line 187: nit: I would invert this to: 
```
if ExactEquals(...) && ExactEquals(...) { 
    return 
}
t.Errorf(...)
```

internal/cmprefimpl/cmppg/checks.go
```diff
@@ -260,19 +260,15 @@ func checkDimension(t *testing.T, want UnaryResult, g geom.Geometry) {
...
 func checkEnvelope(t *testing.T, want UnaryResult, g geom.Geometry) {
 	t.Run("CheckEnvelope", func(t *testing.T) {
-		if g.IsEmpty() {
-			// PostGIS allows envelopes on empty geometries, but they are empty
-			// envelopes. In simplefeatures, an envelope is never empty, so we
-			// skip testing that case.
+		got := g.Envelope().AsGeometry()
+		want := want.Envelope
+
+		// The geometry type for empty envelopers is different for
```
albertteoh, line 266: Not sure if this is what you meant.
```suggestion
		// The geometry type for empty envelopes is different for
```

CHANGELOG.md
peterstace: Good suggestion, fixed in 1e47e80742c5cce93476196eaa3a0be051c1c3de

internal/cmprefimpl/cmppg/checks.go
peterstace: Ahh, yes that's what I meant. Fixed in 1e47e80742c5cce93476196eaa3a0be051c1c3de

geom/type_envelope.go
peterstace: It's mainly to do with what the `Envelope` zero value ends up being. All of the zero values of the other types represent the "empty" variant, so I think it's a good idea to do that here too for consistency. If we were to use `empty` instead of `full` as the field, then the zero value would be the single-point envelope at `(0, 0)`.

peterstace: I wonder if "full" is the wrong word here to best describe the state though?

The _best_ description of what the field really means is "non-empty", although IMO any logic involving boolean negations when the name of the boolean itself contains a negation can be pretty confusing...

geom/util_test.go
peterstace: Good suggestion, that's much easier to read. Changed in 3d837b732f6c1083fdb7857da3f526f8b30c2d8c

geom/type_envelope.go
albertteoh: > All of the zero values of the other types represent the "empty" variant, so I think it's a good idea to do that here too for consistency.

Ah, this is a great reason to use `full` and I think is meaningful enough to easily understand its intent; although it very subtly implies that there are three states, instead of a binary value: 
- full
- not full:
  - empty
  - partially full

> I wonder if "full" is the wrong word here to best describe the state though?

Just brainstorming some other words if you're still looking to rename it: `set`, `isSet`, `initialised`, `isInitialised` (or its American variant), `populated`, `isPopulated`.

Or another alternative is to define an enum of these two values that makes it clearer:
```
const (
   full = iota
   empty
)
```

albertteoh (commented):
> There are some regressions in multipolygon validation performance

What is the root cause of the 15-30% MultiPolygon performance regression? Is there something we can do or it's unavoidable?

peterstace:
> What is the root cause of the 15-30% MultiPolygon performance regression? Is there something we can do or it's unavoidable?

I suspect it's because the Envelope type is going from 32 bytes (4x float64s) to 40 bytes (4x float64s, 1x bool, 7x padding bytes). We're probably now using additional cache lines and might be spilling out somewhere to the next cache level. I don't know that for sure, I just suspect it. If that is the case though, we could possibly step around it by somehow encoding the Envelope (including  the `full` bool) into just 32 bytes. There are a few tricks to do that (they unfortunately all come at the cost of code readability). I'll try them out and see if it makes much difference.

peterstace:
> The other question I forgot to ask is how much attention should be placed on performance and does it vary by package or struct? For example, was it okay to have a 15-30% regression in MultiPolygon validation? The reason why I ask is if in future, other contributors, like myself, add changes and encounter a similar situation where performance regressions are introduced, what guidance should they follow with regards to performance.

I think it comes down to a lot of factors, e.g.:

- What's the nature of the change? E.g. if it's a critical bugfix, then a performance regression is more acceptable (maybe an incorrect algorithm is faster than a correct one).
- How important is the thing being measured? E.g. if it's WKT or GeoJSON parsing, then that's slow anyway since it's string processing, so a performance regression matters less.
- How much harder will the code be to maintain if the regression is fixed? Every performance enhancement that we add (to a varying degree) makes the code harder to reason about and modify. If we have to add a _lot_ of complexity to get a small performance gain, it's probably not worth it.

Generally though, I think we should investigate any _unexpected_ performance regression to at least understand why it's occurring. It's hard to specify exact numbers, but my gut feel is that any regression of > 10% is worthwhile investigating.

peterstace:
Thanks for the review!

PR #421, opened 2021-10-27
Performance improvements to WKB unmarshalling

missinglink:
as discussed in https://github.com/peterstace/simplefeatures/issues/419 this PR improves the Unmarshalling performance of `WKB` by avoiding the `binary.ByteOrder` interface.

this PR itself is purely cosmetic, so I didn't add/change any testing code. it relies on the existing tests to confirm it's a no-op refactor.

I will follow up with a subsequent PR to commit the benchmarking script, that benchmark code is actually more complex than this 😆 so I decided to decouple it from the 'actual work' here.

Overall it shows a marked improvement with the exception of `201ns ± 1%     214ns ± 6%   +6.49%` which is so fast anyway that it's probably just a benchmark fart.

```bash
benchstat -delta-test ttest master.bench pr.bench
name           old time/op    new time/op    delta
WKBParse/0-8      316ns ± 0%     303ns ± 1%   -4.18%  (p=0.000 n=5+5)
WKBParse/1-8      322ns ± 0%     311ns ± 2%   -3.32%  (p=0.005 n=5+5)
WKBParse/2-8      387ns ± 4%     364ns ± 0%   -5.91%  (p=0.008 n=5+5)
WKBParse/3-8      390ns ± 3%     365ns ± 0%   -6.55%  (p=0.002 n=5+5)
WKBParse/4-8      491ns ± 5%     458ns ± 0%   -6.79%  (p=0.006 n=5+5)
WKBParse/5-8      355ns ± 6%     335ns ± 4%     ~     (p=0.058 n=5+5)
WKBParse/6-8      410ns ± 3%     384ns ± 0%   -6.29%  (p=0.002 n=5+4)
WKBParse/7-8      412ns ± 3%     387ns ± 1%   -6.06%  (p=0.001 n=5+5)
WKBParse/8-8      513ns ± 3%     485ns ± 0%   -5.42%  (p=0.005 n=5+5)
WKBParse/9-8      230ns ± 3%     216ns ± 0%   -5.81%  (p=0.004 n=5+4)
WKBParse/10-8     232ns ± 2%     217ns ± 0%   -6.63%  (p=0.000 n=5+5)
WKBParse/11-8     230ns ± 3%     218ns ± 2%   -4.95%  (p=0.008 n=5+5)
WKBParse/12-8     233ns ± 3%     224ns ± 6%     ~     (p=0.056 n=5+5)
WKBParse/13-8     544ns ± 1%     539ns ± 0%     ~     (p=0.056 n=5+5)
[...]

PR #422, opened 2021-10-27
wkb: add benchmarking script

missinglink:
as mentioned in https://github.com/peterstace/simplefeatures/pull/421, this PR adds the benchmarking script used to generate the diffs shown on that PR description.

I'm opening this as a DRAFT since I'm not happy with the current state of this and would like some feedback about how to make it better before opening it up for merging:
- the test fixtures are a copy->paste of the ones from `wkb_test.go`, how best to share these?
- the `decode` function is a copy of `hexStringToBytes` with the `t *testing.T` removed

I tried a few things but none felt quite right:
- create a `wkb_util_test.go` file which has the hex decoder and the PostGIS/Spatialite fixtures in it
- create a `testdata/xxx` file which could be shared.

What do you think? what's the best way to share these between the `wkb_test.go` and `wkb_bench_test.go` files?

Usage:

```bash
go test -bench=. -count=5 -test.run=none -benchmem geom/wkb_bench_test.go > master.bench

# switch branch
go test -bench=. -count=5 -test.run=none -benchmem geom/wkb_bench_test.go > pr.bench

# note the count=5 above, the benchstat command requires each file 
# contains the concatenated output of several runs.
benchstat master.bench pr.bench
```


missinglink:
Hi @peterstace sorry for the long delay, this should be good to go now.

I wasn't able to figure out a nice way of getting the call to `hexStringToBytes()` outside the benchmark codepath, although I'm not super concerned as it will be equal between runs.

One solution would be to add a `[]byte` field to `wkbWKTPair`, let me know if you'd like to go down that route.

missinglink:
In regards to code organization, are you happy with this `geom/wkb_bench_test.go` file or would you prefer to move the `BenchmarkWKBParse` into `geom/wkb_test.go`?

I know some people prefer to have their benchmarks separated from the unit tests, while others prefer to keep them together, I'm happy to follow whatever convention you prefer.

peterstace:
> although I'm not super concerned as it will be equal between runs.

Agreed, I'm not concerned either. Any differences in the benchmarked functionality would still show through as changes since the overhead would be unchanged.

> In regards to code organization, are you happy with this

I'm happy with how it is -- _most_ of the benchmarks are in separate files to regular unit tests, so that fits in well with what you've done.

peterstace (approved):
Thanks for the contribution!

Issue #403, opened 2021-08-11
Implement the json.Unmarshaller interface for each concrete geometry


peterstace:
Currently, the `json.Unmarshaller` interface is only satisfied by the `Geometry` and `GeometryCollection` types.

The concrete geometry types (`Point`, `LineString` etc) should also implement this interface.

peterstace:
We should do a similar thing for the SQL related methods as well.

peterstace:
We need to be careful here to ensure that geometries remain immutable after creation (with unit tests to show it).

peterstace:
> We should do a similar thing for the SQL related methods as well.

This part was already completed in https://github.com/peterstace/simplefeatures/issues/351

PR #426, opened 2021-11-02
Fix Spelling Error In Marshaler Comment

peterstace:
## Description

Fix "marshaler" spelling to match the Go std lib.

It seems as though there are two spellings of "Marshal/Marshall" and "Marshaller/Marshaller". When referring to the `encoding/json.Marshaler` interface, we should use the same spelling as the standard library (since they [define that interface](https://pkg.go.dev/encoding/json#Marshaler)). For consistency, we should use the same spelling in other places too.

## Check List

Have you:

- Added unit tests? N/A

## Related Issue

Noticed as part of https://github.com/peterstace/simplefeatures/issues/403

## Benchmark Results

N/A

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## Unreleased
+
+- Fix spelling of "Marshaller" when referring to the interface defined in the
+  `encoding/json` package.
+
 ## v0.34.0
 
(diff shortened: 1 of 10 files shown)


albertteoh (approved):

peterstace:
Thanks for the review!

PR #427, opened 2021-11-07
Unmarshal GeoJSON into Concrete Geometries

peterstace:
## Description

- Adds `UnmarshalJSON` methods to each concrete geometry type.

- This means that users will be able to unmarshal GeoJSON directly into a concrete geometry type if they know it's of a particular concrete type. Previously, they would have to unmarshal into the generic `Geometry` type, and then check to see if it's actually the type they expected (and extract it).

## Check List

Have you:

- Added unit tests? Yes.

## Related Issue

- https://github.com/peterstace/simplefeatures/issues/403

## Benchmark Results

N/A

### DIFF
--- a/geom/type_geometry_collection.go
+++ b/geom/type_geometry_collection.go
@@ -212,6 +212,24 @@ func (c GeometryCollection) MarshalJSON() ([]byte, error) {
 	return buf, nil
 }
 
+// UnmarshalJSON implements the encoding/json.Unmarshaler interface by decoding
+// the GeoJSON representation of a GeometryCollection.
+func (c *GeometryCollection) UnmarshalJSON(buf []byte) error {
+	g, err := UnmarshalGeoJSON(buf)
+	if err != nil {
+		return err
+	}
+	gc, ok := g.AsGeometryCollection()
+	if !ok {
+		return wrongTypeDuringUnmarshalError{
+			destType:   TypeGeometryCollection,
+			actualType: g.Type(),
+		}
+	}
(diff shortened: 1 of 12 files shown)


geom/type_geometry_collection.go
```diff
@@ -212,6 +212,24 @@ func (c GeometryCollection) MarshalJSON() ([]byte, error) {
 	return buf, nil
 }
 
+// UnmarshalJSON implements the encoding/json.Unmarshaler interface by decoding
```
peterstace, line 215: These implementations are pretty boilerplatey... I don't see an easy way around it though.

peterstace: I tried using reflect to remove the boilerplate (the approach works -- unit tests still pass). However, it's pretty hard to read, and overall I _think_ it's probably not an improvement even though it does successfully eliminate 90% of the boilerplate. I've reverted the change, but would like to get anyone's opinion on it. 7d07b3852658d81e3804113e761c1ccf1faddaf3

peterstace: I've got a new approach to reduce boilerplate here: a855eacfeec8029626b9988282f9c251b5db46dd It follows the same strategy taken for WKB scanning boilerplate (and even shares some of the type dispatch code). I think that's a bit cleaner, and doesn't rely on reflect.

albertteoh (approved):
LGTM 👍🏼

peterstace:
Thanks for reviewing! 😁

Issue #416, opened 2021-09-08
Wrap errors using the `%w` format verb


peterstace:
The `wrap` function in `geom/errors.go` currently uses the `%v` verb to wrap errors opaquely. We don't actually need opaque wrapping though (since concrete error types such as `validationError` are not named with exported symbols).

See [here](https://github.com/peterstace/simplefeatures/pull/406#discussion_r704313254) for full context.

Steps:

- Use the `%w` verb rather than the `%v` verb in the `wrap` implementation.

PR #428, opened 2021-11-11
issue-416 use %w when wrapping errors

sameeraaperera:
## Description

Use usage of %v to %w when wrapping errors so that it will become available to `errors.Is` and `errors.As` checks.
Found some other places outside geom/errors.go where this was being used and updated those too.

## Check List

Have you:

- Added unit tests? N/A

## Related Issue

- https://github.com/peterstace/simplefeatures/issues/416

## Benchmark Results

N/A


peterstace (approved):
Looks good! Thank you for the contribution 😁

Issue #425, opened 2021-11-02
Allow concrete geometries to be simplified


peterstace:
Simplify functionality is currently exposed via a package-level function with signature `func Simplify(g Geometry, threshold float64, opts ...ConstructorOption) (Geometry, error)`.

This makes the API simple in the sense that it has a small surface area. However, it's not type-safe for some usecases.

For example, if a user has a `Polygon` and wants to simplify it (and get back a `Polygon`), but must use a construct such as:

```
var poly geom.Polygon
// populate poly
result, err := geom.Simplify(poly.AsGeometry(), 0.5)
// check error
if simpPoly, ok := result.AsPolygon(); ok {
    // can now do stuff with simpPoly, which is of type Polygon.
    // Alternatively, could use MustAsPolygon to avoid the ok check.
}
```

The `Simplify` package-level function always returns a `Geometry` that has the same `Type` as its input, so using `MustAsPolygon` would be safe in practice. However, that guarantee is not documented, so it's not necessarily a good idea for users to do that.

In order to give better type safety, we could do the following:

- Add new methods to `LineString`, `MultiLineString`, `Polygon`, `MultiPolygon`, and `GeometryCollection`. The method names should all be `Simplify`, and have method signature `Simplify(threshold float64, opts ...ConstructorOption) (T, error)` (where `T` is the Go type that the method is defined on). It doesn't make much sense to define a `Simplify` method on `Point` or `MultiPoint`, since they would be no-ops.

- Change the package level-function to a method on the `Geometry` type for consistency. This would be a breaking change, but one that would be easy for users to adapt to.

PR #431, opened 2021-11-15
Refactor simplify to be a method rather than a function

peterstace:
## Description

- Moves the `Simplify` package level function to be a method on the `Geometry` type.

- Adds `Simplify` methods to `LineString`, `MultiLineString`, `Polygon`, `MultiPolygon`, 
and `GeometryCollection`. These methods return the same concrete geometry type as the type the methods are on, which improves type safety in some scenarios.

- Fixes a bug in polygon simplification.

- See linked issue for motivating example.

## Check List

Have you:

- Added unit tests? Yes.

## Related Issue

- https://github.com/peterstace/simplefeatures/issues/425

## Benchmark Results

N/A

### DIFF
--- a/geom/alg_simplify_test.go
+++ b/geom/alg_simplify_test.go
@@ -123,6 +123,18 @@ func TestSimplifyErrorCases(t *testing.T) {
 			))`,
 			1e-5,
 		},
+
+		// Second case for "outer ring becomes invalid after simplification".
+		{`POLYGON((0 0,0 1,0.9 1,1 1.1,1.1 1,2 1,2 0,1 1.05,0 0))`, 0.2},
+
+		// Inner ring becomes invalid after simplification.
+		{
+			`POLYGON(
+				(-1 -1,-1 3,3 3,3 -1,-1 -1),
+				(0 0,0 1,0.9 1,1 1.1,1.1 1,2 1,2 0,1 1.05,0 0)
+			)`,
+			0.2,
+		},
 	} {
 		t.Run(strconv.Itoa(i), func(t *testing.T) {
 			in := geomFromWKT(t, tc.wkt)
--- a/geom/alg_simplify_test.go
+++ b/geom/alg_simplify_test.go
@@ -125,6 +125,9 @@ func TestSimplifyErrorCases(t *testing.T) {
 		},
 
 		// Second case for "outer ring becomes invalid after simplification".
+		// The outer ring becomes invalid because the Ramer-Douglas-Peucker
+		// algorithm causes the ring (when considered as a LineString) to
+		// become self-intersected.
 		{`POLYGON((0 0,0 1,0.9 1,1 1.1,1.1 1,2 1,2 0,1 1.05,0 0))`, 0.2},
 
 		// Inner ring becomes invalid after simplification.
--- a/geom/alg_simplify_test.go
+++ b/geom/alg_simplify_test.go
@@ -95,7 +95,7 @@ func TestSimplify(t *testing.T) {
 			t.Logf("input:     %v", in.AsText())
 			t.Logf("threshold: %v", tc.threshold)
 			t.Logf("want:      %v", want.AsText())
-			got, err := geom.Simplify(in, tc.threshold)
+			got, err := in.Simplify(tc.threshold)
 			expectNoErr(t, err)
 			t.Logf("got:       %v", got.AsText())
 			expectGeomEq(t, got, want, geom.IgnoreOrder)
@@ -126,7 +126,7 @@ func TestSimplifyErrorCases(t *testing.T) {
 	} {
 		t.Run(strconv.Itoa(i), func(t *testing.T) {
 			in := geomFromWKT(t, tc.wkt)
-			_, err := geom.Simplify(in, tc.threshold)
+			_, err := in.Simplify(tc.threshold)
 			expectErr(t, err)
 		})
 	}
--- a/geom/type_line_string.go
+++ b/geom/type_line_string.go
@@ -39,7 +39,6 @@ func NewLineString(seq Sequence, opts ...ConstructorOption) (LineString, error) 
 
 func newLineStringWithOmitInvalid(seq Sequence) LineString {
 	if err := validateLineStringSeq(seq); err != nil {
-		// TODO: Should the coordinates type match the input seq?
 		return LineString{}
 	}
 	return LineString{seq}
--- a/geom/type_line_string.go
+++ b/geom/type_line_string.go
@@ -32,6 +32,7 @@ func NewLineString(seq Sequence, opts ...ConstructorOption) (LineString, error) 
 	// Valid non-empty LineStrings must have at least 2 *distinct* points.
 	if !hasAtLeast2DistinctPointsInSeq(seq) {
 		if ctorOpts.omitInvalid {
+			// TODO: Should the coordinates type match the input seq?
 			return LineString{}, nil
 		}
 		return LineString{}, validationError{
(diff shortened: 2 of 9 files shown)


geom/type_polygon.go
```diff
@@ -592,3 +592,40 @@ func (p Polygon) Summary() string {
...
+	if hasCollapsed(exterior) {
+		return Polygon{}, nil
+	}
+
+	n := p.NumInteriorRings()
+	rings := make([]LineString, 0, n+1)
+	rings = append(rings, exterior)
+	for i := 0; i < n; i++ {
+		interior := p.InteriorRingN(i).Simplify(threshold)
+		if !hasCollapsed(interior) {
```
peterstace, line 625: The bug fix mentioned in the PR description is here. Previously, the condition on this if statement was `!interior.IsRing` and was changed to `!hasCollapsed`. The implication of the bug was that if an interior ring became invalid but _didn't_ collapse, we would omit it from the output. Instead, we now give an error (which matches GEOS and PostGIS behaviour).

geom/type_polygon.go
```diff
@@ -592,3 +592,40 @@ func (p Polygon) Summary() string {
...
+// prohibited by Polygon validation rules (such as intersecting at more than
+// one point). In these cases, an error is returned. Construction behaviour of
+// the output (which includes omitting errors) may be controlled via
+// ConstructorOptions.
+func (p Polygon) Simplify(threshold float64, opts ...ConstructorOption) (Polygon, error) {
+	exterior := p.ExteriorRing().Simplify(threshold)
+
+	// If we don't have at least 4 coordinates, then we can't form a ring, and
+	// the polygon has collapsed either to a point or a single linear element.
+	// Both cases are represented by an empty polygon.
```
albertteoh, line 612: Nice comment 👍🏼 

nit: for consistency, I noticed we used `Polygon` in other comments.

```suggestion
	// Both cases are represented by an empty Polygon.
```

geom/alg_simplify_test.go
```diff
@@ -123,10 +123,22 @@ func TestSimplifyErrorCases(t *testing.T) {
 			))`,
 			1e-5,
 		},
+
+		// Second case for "outer ring becomes invalid after simplification".
+		{`POLYGON((0 0,0 1,0.9 1,1 1.1,1.1 1,2 1,2 0,1 1.05,0 0))`, 0.2},
+
+		// Inner ring becomes invalid after simplification.
```
albertteoh, line 130: If we know the reason for the inner ring becoming invalid, do you think it would be useful to add this reason in this comment?

geom/type_line_string.go
```diff
@@ -32,6 +32,7 @@ func NewLineString(seq Sequence, opts ...ConstructorOption) (LineString, error)
 	// Valid non-empty LineStrings must have at least 2 *distinct* points.
 	if !hasAtLeast2DistinctPointsInSeq(seq) {
 		if ctorOpts.omitInvalid {
+			// TODO: Should the coordinates type match the input seq?
```
albertteoh, line 35: For my understanding, is this TODO relating to the empty `LineString{}` being returned?

geom/type_line_string.go
```diff
@@ -422,3 +423,19 @@ func (s LineString) Summary() string {
 func (s LineString) String() string {
 	return s.Summary()
 }
+
+// Simplify returns a simplified version of the LineString using the
+// Ramer-Douglas-Peucker algorithm. If the Ramer-Douglas-Peucker were to create
```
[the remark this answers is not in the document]
peterstace: Ah, that was unintentional.

geom/type_line_string.go
```diff
@@ -422,3 +423,19 @@ func (s LineString) Summary() string {
...
+// an invalid LineString (i.e. one having only a single distinct point), then
+// the empty LineString is returned.
+func (s LineString) Simplify(threshold float64) LineString {
+	seq := s.Coordinates()
+	floats := ramerDouglasPeucker(nil, seq, threshold)
+	seq = NewSequence(floats, seq.CoordinatesType())
+
+	simp, err := NewLineString(seq, OmitInvalid)
+	if err != nil {
+		panic("OmitInvalid used, so cannot panic: " + err.Error())
```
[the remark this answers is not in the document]
peterstace: Agreed, `"OmitInvalid used, unexpected error: "` is better here.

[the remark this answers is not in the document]
peterstace: > Is it because it's unexpected and should never happen?

Yes, that's right. Since it should never happen, it seems it seems unkind to users to require them to check an error. Perhaps there is a way to refactor things such that we don't have to panic on an error that should never happen, we might be able to add some internal LineString constructors that don't give an error. Leave it with me, I'll have a think of some better ways to do this.

geom/alg_simplify_test.go
peterstace: Agreed, I'll add that.

geom/type_line_string.go
peterstace: Yes, that's correct. The TODO is not related to this PR at all, so I'll create a new GH issue and remove the TODO.

geom/type_line_string.go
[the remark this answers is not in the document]
peterstace: I've made a change to improve this in bc576058cf530e201d7e6dc507c5bfae423df2d2

It basically creates a new unexported `LineString` constructor that implicitly uses the `OmitEmpty` behaviour, and therefore _doesn't_ need to return an error. We can then call that directly from the `Simplify` method.

geom/alg_simplify_test.go
peterstace: I've added to the comment in ca0e2de732356b70b26aea2b546fff34a6074fe0, but I'm not sure it makes it _completely_ clear from the comment itself. I think ultimately with these sorts of tests, a reader (or a developer debugging!) would really need to view the polygons visually to really understand what's going on. In this case, the polygon with inner ring looks like this:

Screen Shot 2021-11-17 at 5 46 35 am

The simplify algorithm would remove a single control point (the most northern point in the inner ring). That causes the inner ring to overlap on itself.

geom/type_line_string.go
peterstace: I've created a new issue to track this problem: https://github.com/peterstace/simplefeatures/issues/432

I've also removed the TODO in ee74938511f5b9a047318b61a19b68fc1ea78b0e (since it's tracked by an issue, the TODO just clutters things up).

geom/alg_simplify_test.go
albertteoh: I agree that it's hard to really understand without an image, but I like that it mentions that executing the Simplify operation causes an inner ring to intersect with itself.

Although, if we remove the northern point in the inner ring, I don't understand how the inner ring would overlap on itself. I thought it would just result in a line string within the polygon since the inner ring is no longer connected?

albertteoh (approved):
LGTM!

geom/alg_simplify_test.go
peterstace: > Although, if we remove the northern point in the inner ring, I don't understand how the inner ring would overlap on itself. I thought it would just result in a line string within the polygon since the inner ring is no longer connected?

So when I say "remove the northern most control point", I mean that the two control points adjacent to it will join up ("skipping" the middle one). An image of the geometry in that invalid geometry is below:

Screen Shot 2021-11-17 at 7 51 30 am

albertteoh: Aaaah, I see, it's the implicit re-joining that I wasn't aware of. Thanks!

PR #433, opened 2021-11-17
GEOS: Wrap Make Valid

peterstace:
## Description

- Wraps the `GEOSMakeValid_r` function from GEOS.

## Check List

Have you:

- Added unit tests? Yes, just to confirm that the wrapping is working. Doesn't actually do extensive tests of the functionality.

## Related Issue

- https://github.com/peterstace/simplefeatures/issues/430

## Benchmark Results

N/A

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,9 @@ __Special thanks to Albert Teoh and Sameera Perera for contributing to this rele
 - Fixes a bug in Simplify where invalid interior rings would be omitted rather
   than producing an error.
 
+- Adds a wrapper in the `geos` package for the `GEOSMakeValid_r` function
+  (exposed as `MakeValid`).
+
 ## v0.34.0
 
 2021-11-02
--- a/geos/entrypoints.go
+++ b/geos/entrypoints.go
@@ -11,7 +11,7 @@ package geos
 	(GEOS_VERSION_MAJOR == 3 && GEOS_VERSION_MINOR < 8) \
 )
 #if MAKE_VALID_MISSING
-// This stub implementation also fails
+// This stub implementation always fails:
 GEOSGeometry *GEOSMakeValid_r(GEOSContextHandle_t handle, const GEOSGeometry* g) { return NULL; }
 #endif
 
--- a/geos/entrypoints.go
+++ b/geos/entrypoints.go
@@ -309,3 +309,15 @@ func SymmetricDifference(a, b geom.Geometry, opts ...geom.ConstructorOption) (ge
 	})
 	return result, wrap(err, "executing GEOSSymDifference_r")
 }
+
+// MakeValid can be used to convert an invalid geometry into a valid geometry.
+// It does this by keeping the original control points and constructing a new
+// geometry that is valid and similar (but not the same as) the original
+// invalid geometry. If the input geometry is valid, then it is returned
+// unaltered.
+func MakeValid(g geom.Geometry, opts ...geom.ConstructorOption) (geom.Geometry, error) {
+	result, err := unaryOpG(g, opts, func(ctx C.GEOSContextHandle_t, g *C.GEOSGeometry) *C.GEOSGeometry {
+		return C.GEOSMakeValid_r(ctx, g)
+	})
+	return result, wrap(err, "executing GEOSMakeValid_r")
+}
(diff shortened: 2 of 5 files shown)


peterstace:
I think I've got a decent solution for this in c8e5fa0c16f350a8cf8d2019cbd5912f7d74ac7c

It basically checks the current GEOS version, and defines a C stub if it doesn't exist. It then checks the version and gives a friendly error if it's about to call the stub.

geos/handle.go
```diff
@@ -62,6 +61,7 @@ func newHandle() (*handle, error) {
 		h.release()
 		return nil, errors.New("malloc failed")
 	}
+	C.memset((unsafe.Pointer)(h.errBuf), 0, 1024)
```
peterstace, line 64: This was a bug I noticed while I was here. Malloc is not going to actually zero out the memory content... (I've been spoilt by Go).

geos/entrypoints_test.go
```diff
@@ -767,3 +767,26 @@ func TestSymmetricDifference(t *testing.T) {
...
+		input      string
+		wantOutput string
+	}{
+		{
+			"POLYGON((0 0,2 2,2 0,0 2,0 0))",
+			"MULTIPOLYGON(((0 0,0 2,1 1,0 0)),((2 0,1 1,2 2,2 0)))",
+		},
+	} {
+		t.Run(strconv.Itoa(i), func(t *testing.T) {
+			in := geomFromWKT(t, tt.input, geom.DisableAllValidations)
```
albertteoh, line 782: Is there a way to assert that the `input` geometry is invalid?

CHANGELOG.md
```diff
@@ -30,6 +30,9 @@ __Special thanks to Albert Teoh and Sameera Perera for contributing to this rele
 - Fixes a bug in Simplify where invalid interior rings would be omitted rather
   than producing an error.
 
+- Adds a wrapper in the `geos` package for the `GEOSMakeValid_r` function
```
albertteoh, line 33: What was the motivation for wrapping the `GEOSMakeValid_r` function? What challenges are there to implementing this natively in Go?

geos/entrypoints.go
```diff
@@ -4,6 +4,17 @@ package geos
...
 #include "geos_c.h"
+
+#define MAKE_VALID_MIN_VERSION "3.8.0"
+#define MAKE_VALID_MISSING ( \
+	GEOS_VERSION_MAJOR < 3 || \
+	(GEOS_VERSION_MAJOR == 3 && GEOS_VERSION_MINOR < 8) \
+)
+#if MAKE_VALID_MISSING
+// This stub implementation always fails:
+GEOSGeometry *GEOSMakeValid_r(GEOSContextHandle_t handle, const GEOSGeometry* g) { return NULL; }
```
albertteoh, line 15: Is this needed to prevent a compile-time error?

albertteoh (commented):
Nice

geos/entrypoints_test.go
peterstace: Yes, that's possible. Added in 65c283780294dd28c604144e5a1b32be7a024a28

CHANGELOG.md
peterstace: It's something that Ken _may_ find useful in his current work (might be better to take that discussion offline for further details).

The main challenges involved there would be coming up with an actual algorithm to do it (possibly that's the hard part) and the implementing it. It would most likely involve a [DCEL](https://en.wikipedia.org/wiki/Doubly_connected_edge_list) data structure (which we have already implemented). I'd ballpark estimate 50hrs to 100hrs of effort would be required.

geos/entrypoints.go
peterstace: Yes, that's correct. Prior versions of the GEOS library don't define the symbol at all.

CHANGELOG.md
albertteoh: If there are plans to do this, maybe it's worth creating a ticket for this work; but if it's not in scope of the goals of simplefeatures, then all good.

albertteoh (approved):
LGTM

CHANGELOG.md
peterstace: Good call, it's definitely in scope. I've created a ticket here: https://github.com/peterstace/simplefeatures/issues/435

peterstace:
Thanks for the review, I really appreciate it! 🥳

Issue #436, opened 2021-12-01
Area transform functions are invoked multiple times


peterstace:
In the `signedAreaOfLinearRing` function, the `transform` function is evaluated twice per control point. An obvious performance improvement would be to evaluate only once per control point.

PR #437, opened 2021-12-01
Eliminate redundant transformations in area calculation

peterstace:
## Description

    Reduce number of transform invocations in area calc
    
    When calculating an area with a transformation, we only really have to
    transform each point exactly once. However, the previous implementation
    was naively calculating each transformation twice. For simple
    transformations, this may not matter much. However, the transformation
    is user-supplied, so could be arbitrarily expensive to compute.
    
    The bounds of the for loop in the shoelace formula implementation have
    been reduced by 1 from from `[0, n)` to `[0, n-1)`. This is because the
    shoelace formula applies to rings, and the start and end points of rings
    are always the same. So the final iteration in the previous
    implementation was redundant.

## Check List

Have you:

- Added unit tests? Yes, a new unit test to check the number of invocations. Existing unit tests ensure that the result of the area calculation is still correct.

## Related Issue

- https://github.com/peterstace/simplefeatures/issues/436

## Benchmark Results

N/A

### DIFF
--- a/geom/attr_test.go
+++ b/geom/attr_test.go
@@ -642,6 +642,18 @@ func TestTransformedArea(t *testing.T) {
 	}
 }
 
+func TestTransformedAreaInvocationCount(t *testing.T) {
+	g := geomFromWKT(t, "POLYGON((0 0,0 1,1 0,0 0))")
+	var count int
+	g.Area(WithTransform(func(xy XY) XY {
+		count++
+		return xy
+	}))
+
+	// Each of the 4 points making up the polygon get transformed once each.
+	expectIntEq(t, count, 4)
+}
+
 func TestCentroid(t *testing.T) {
 	for i, tt := range []struct {
 		input  string
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## Unreleased
+
+- Eliminate redundant calls to the optional user supplied transform func during
+  area calculations.
+
 ## v0.35.0
 
 2021-11-23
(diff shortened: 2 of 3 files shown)


geom/attr_test.go
```diff
@@ -642,6 +642,18 @@ func TestTransformedArea(t *testing.T) {
...
+func TestTransformedAreaInvocationCount(t *testing.T) {
+	g := geomFromWKT(t, "POLYGON((0 0,0 1,1 0,0 0))")
+	var count int
+	g.Area(WithTransform(func(xy XY) XY {
+		count++
+		return xy
+	}))
+
+	// Each of the 4 points making up the polygon get transformed once each.
+	expectIntEq(t, count, 4)
```
albertteoh, line 654: I've probably misunderstood something, but in my mind, would have expected `count` to be 3 since the POLYGON is a triangle (3 vertices) made up of 4 points and we're looping 4-1=3 times?

albertteoh (approved):

albertteoh:
LGTM, and I'm sure I'm wrong, but just wanted to clarify why my understanding is wrong. :)

geom/attr_test.go
peterstace: You're on the right track, but I think you may have missed that `nthPt` is called once _before_ the loop starts. So yes, we do loop `n-1` times, but we do `1` transformation outside the loop as well.

albertteoh: Ah yes, the `pt1 := nthPt(0)`. Thanks for clarifying!

albertteoh (approved):
LGTM

peterstace:
Thanks for the review!

Issue #382, opened 2021-07-17
Add IsPolygonCW and IsPolygonCCW


peterstace:
Add new methods `IsCW() bool` and `IsCCW` to `Polygon`, `MultiPolygon`, `GeometryCollection`, and `Geometry`.

They should return true if and only if each polygon contained in the geometry, the windings are CW for the outer ring and CCW for the inner rings in the case of `IsCW`. I.e. the designation CW in the method name refers to the orientation of the outer ring, and the opposite orientation is implied for inner rings.

For `IsCCW`, the check is performed but with orientations reversed.

peterstace:
Fixed in #438

PR #438, opened 2022-01-21
Add IsCW, IsCCW to Polygon, MultiPolygon, Geometry, GeometryCollection

DoctorLoki:
## Description

Adds IsCW and IsCCW methods.

Note I have defined certain behaviours for empty and inconsistent geometries.
In particular, empty and zero area rings are defined to be neither CW nor CCW.

Note the above assumption may differ from existing usage where it is assumed !CW == CCW.
See e.g. https://github.com/peterstace/simplefeatures/blob/master/geom/type_polygon.go#L553
It is unclear if changes are needed in such locations to be more consistent with the above assumption.

## Check List

Have you:

- Added unit tests? yes

- Add cmprefimpl tests? (if appropriate?) no

- Updated release notes? (if appropriate?) no

## Related Issue

https://github.com/peterstace/simplefeatures/issues/382


albertteoh (approved):
The high level logic LGTM, but I'll defer to @peterstace because the `signedAreaOfLinearRing` part is beyond me 😅 .

DoctorLoki:
I'm using `signedAreaOfLinearRing` to provide the definition of CW, as it has been used elsewhere for this purpose.

I considered adding `linearRingIsCW` and `linearRingIsCCW` methods to even further abstract that logic, but ultimately thought that too much. But if that is considered a good idea, I could do that.

peterstace:
Thanks for the contribution! 😁 🍰 

The code itself and the tests look pretty good.

I'm just having a think about the definitional aspects around corner cases (in particular when Polygons/MultiPolygons are empty, when a GeometryCollection contains stuff but _not_ polygons, or when the Geometry is non-areal, e.g. a LineString).

[`ST_IsPolygonCW`](http://postgis.net/docs/ST_IsPolygonCW.html) returns true in the case where a Polygon is empty, or the input is non-areal (e.g. a Point or LineString), or when a GeometryCollection contains no Polygons/MultiPolygons.

That's a change in semantics from "the geometry contains at least 1 polygon and all polygons are oriented CW" (i.e. what you've implemented) to "the geometry as a whole is consistent with CW orientation" (i.e. what PostGIS implements). To me, the latter seems a bit more useful. I imagine the most common usecase for these sorts of methods would be when you want to check whether a geometry you have is consistent with a particular orientation (so that you can force an orientation, or log a warning, or something else, if the geometry is not consistent with the orientation you want).

Note that GEOS doesn't implement this behaviour on geometries, so we can't really use that for inspiration.

I _think_ all you'd need to do to change the behaviour is change to `return true` in the `default` cases of the switches, and remove the `if c.IsEmpty() { return false }` checks at the start of each method.

Thoughts?

DoctorLoki:
Could certainly make that change (it's close to what I originally implemented until I ran some tests and thought about it more). The naming feels slightly wrong, but I agree that's what you'd typically want to use these methods for.

peterstace (approved):
LGTM!

Issue #54, opened 2019-09-04
Support for TWKB (Tiny Well Known Binary)


den3tsou:
The spec is [this](https://github.com/TWKB/Specification/blob/master/twkb.md)

peterstace:
Fixed in #441

PR #441, opened 2022-02-06
Add Tiny Well-Known Binary decoder and encoder.

DoctorLoki:
## Description

Add Tiny Well-Known Binary decoding and encoding capability with unit tests.

## Check List

Have you:

- Added unit tests? Yes

- Add cmprefimpl tests? No

- Updated release notes? No

## Related Issue

Implements https://github.com/peterstace/simplefeatures/issues/54


geom/twkb_parser.go
```diff
@@ -0,0 +1,554 @@
...
+		return Polygon{}, p.parserError("cannot unmarshal num rings varint malformed")
+	}
+
+	var rings []LineString
+	for r := uint64(0); r < numRings; r++ {
+		coords, err := p.parsePointCountAndArray()
+		if err != nil {
+			return Polygon{}, err
+		}
+		if len(coords) > 0 {
```
DoctorLoki, line 313: This part is of concern to me.
See multiple polygon examples at https://github.com/TWKB/twkb.js/blob/master/test/twkb.spec.js#L43 which seem to do things differently, and particularly https://github.com/TWKB/twkb.js/blob/master/test/twkb.spec.js#L169 but compare that with the spec and you can see why I'm concerned.

geom/twkb_parser.go
```diff
@@ -0,0 +1,554 @@
...
+		return p.parserError("cannot unmarshal lacking metadata header")
+	}
+	metaheader := p.twkb[p.pos]
+	p.pos++
+
+	p.hasBBox = (metaheader & 1) != 0
+	p.hasSize = (metaheader & 2) != 0
+	p.hasIDs = (metaheader & 4) != 0
+	p.hasExt = (metaheader & 8) != 0
+	p.isEmpty = (metaheader & 16) != 0
```
DoctorLoki, line 178: The spec seems unclear about whether any further data is needed if the metadata header byte indicated the object is empty. In my tests I have assumed no extra bytes are required. This may b at variance from the reference implementation https://github.com/TWKB/twkb.js/blob/master/src/readBuffer.js which sets `is_empty` but then ignores the field. So I may need to change the tests and implementation to also require a zero number of points/rings/etc to be present.

DoctorLoki: Update: PostGIS seems to agree with my implementation:
```
=> select ST_AsTWKB('POINT EMPTY'::geometry);
 st_astwkb 
-----------
 \x0110

surveys=> select ST_AsTWKB('POLYGON EMPTY'::geometry);
 st_astwkb 
-----------
 \x0310
```

geom/twkb_parser.go
DoctorLoki: It seems that at least one implementation emits closed rings when encoding (note there are 5 points given for each ring):
```
=> select ST_AsTWKB('POLYGON((0 0,2 0,2 2,0 2,0 0),(0 0,0 1,1 1,1 0,0 0))'::geometry);
                      st_astwkb                       
------------------------------------------------------
 \x03000205000004000004030000030500000002020000010100
```
When decoding, both closed and unclosed rings are accepted and correctly handled as below:
```
=> select ST_AsText(ST_GeomFromTWKB(E'\\x03000205000004000004030000030500000002020000010100'));
                      st_astext                       
------------------------------------------------------
 POLYGON((0 0,2 0,2 2,0 2,0 0),(0 0,0 1,1 1,1 0,0 0))

=> select ST_AsText(ST_GeomFromTWKB(E'\\x030002040000040000040300040003000202000001'));
                      st_astext                       
------------------------------------------------------
 POLYGON((0 0,2 0,2 2,0 2,0 0),(0 0,0 1,1 1,1 0,0 0))
```
So I think the parser should handle both closed and unclosed rings.

DoctorLoki: I've now implemented that change, to allow the parser to accept either closed or unclosed rings.

geom/twkb_parser.go
```diff
@@ -0,0 +1,562 @@
...
+	if n <= 0 {
+		return 0, p.parserError("problem parsing unsigned varint")
+	}
+	p.pos += n // Have now read the varint.
+	return val, nil
+}
+
+func (p *TWKBParser) parseSignedVarint() (int64, error) {
+	// LEB128 Zig-Zag encoded.
+	uval, n := binary.Uvarint(p.twkb[p.pos:])
```
albertteoh, line 520: Did you mean to use `binary.Varint`?

geom/twkb_parser.go
```diff
@@ -0,0 +1,562 @@
...
+			// The spec says rings may need to be closed:
+			// "rings are assumed to be implicitly closed, so the first
+			//  and last point should not be the same"
+
+			// Note, some implementations can generate TWKB with the rings
+			// already closed. We wish to gracefully parse these cases too.
+			// So determine if the first and final coords differ.
+			finalPointDiffersFromFirst := false
+			for d := 0; d < p.dimensions; d++ {
+				first := coords[0+d]
```
albertteoh, line 323: What's the reason for `0+d` instead of just `d`?

geom/twkb_parser.go
```diff
@@ -0,0 +1,562 @@
...
+	}
+	p.pos += n // Have now read the varint.
+	return val, nil
+}
+
+func (p *TWKBParser) parseSignedVarint() (int64, error) {
+	// LEB128 Zig-Zag encoded.
+	uval, n := binary.Uvarint(p.twkb[p.pos:])
+	if n <= 0 {
+		return 0, p.parserError("problem parsing signed varint")
```
albertteoh, line 522: The [docs](https://pkg.go.dev/encoding/binary#Varint) differentiate `n == 0` and `n < 0`, which have different meanings. Do you think it's worth conveying these meanings in separate error messages?

```
n == 0: buf too small
n  < 0: value larger than 64 bits (overflow)
        and -n is the number of bytes read
```

geom/twkb_parser.go
```diff
@@ -0,0 +1,562 @@
...
+	p.hasSize = (metaheader & 2) != 0
+	p.hasIDs = (metaheader & 4) != 0
+	p.hasExt = (metaheader & 8) != 0
+	p.isEmpty = (metaheader & 16) != 0
+	return nil
+}
+
+func (p *TWKBParser) parseExtendedPrecision() error {
+	if len(p.twkb) <= p.pos {
+		return p.parserError("cannot unmarshal lacking extended pecision byte")
```
albertteoh, line 184:
```suggestion
		return p.parserError("cannot unmarshal: lacking extended precision byte")
```

albertteoh: nit: I suggest adding a `:` after each `cannot unmarshal` in error messages (same below) as I feel we're annotating two error messages here and it helps with reading the error message:
- The fact that unmarshalling failed.
- The reason why unmarshalling failed.

geom/twkb_parser.go
```diff
@@ -0,0 +1,562 @@
...
+	var polys []Polygon
+	for i := 0; i < int(numPolygons); i++ {
+		poly, err := p.nextPolygon()
+		if err != nil {
+			return MultiPolygon{}, err
+		}
+		polys = append(polys, poly)
+	}
+	mp, err := NewMultiPolygon(polys, p.opts...)
+	return mp, p.annotateError(err)
```
albertteoh, line 427: I noticed that `annotateError` is used in some places but not others. 

Would it make sense to annotate the error once in `UnmarshalTWKB` (bottom of the call stack) and instead of `p.parserError`, return the built-in error via `fmt.Errorf(...)` or `errors.New(...)` instead? It would help reduce potential duplication of the byte position annotation in error messages.

geom/twkb_parser.go
DoctorLoki: Good point, I'll make the error messages a bit clearer.

geom/twkb_parser.go
DoctorLoki: It was just to reinforce the fact we're reading from the first point (at offset 0). It's obviously not strictly necessary. I can remove it.

geom/twkb_parser.go
DoctorLoki: I was trying to use it only in return paths where the byte position wasn't otherwise being reported. Good point about annotating in one spot, I'll try to tidy that up.

geom/twkb_parser.go
DoctorLoki: Good point, `binary.Varint` reverses the zig-zag encoding so I don't need to do it myself. Will change.

geom/twkb_parser.go
albertteoh: Okay, that makes sense, we can leave it there, or maybe put it in an `offset` variable if you feel it helps with readability. No strong opinion either way.

geom/twkb_parser.go
DoctorLoki: I didn't bother differentiating those because they're going to lead to the parse being aborted either way, but you're right, separate error messages could help with debugging.

geom/twkb_parser.go
```diff
@@ -0,0 +1,525 @@
...
+		dimensions: 2,
+	}
+}
+
+// Annotate an error message with the byte offset where it happened.
+func (p *TWKBParser) annotateError(err error) error {
+	if err == nil {
+		return nil
+	}
+	return fmt.Errorf("TWKB parsing error at byte %d: %s", p.pos, err.Error())
```
albertteoh, line 65: The `%w` verb allows errors to be used with `errors.Is` and `errors.As`.

See: https://go.dev/blog/go1.13-errors

```suggestion
	return fmt.Errorf("TWKB parsing error at byte %d: %w", p.pos, err)
```

albertteoh: I noticed a few more cases in this PR where we could use the `%w` verb as well.

DoctorLoki: Nice, I'm using the `%w` verb now.

geom/twkb.go
```diff
@@ -0,0 +1,50 @@
+package geom
+
+// Tiny Well Known Binary
+// See spec https://github.com/TWKB/Specification/blob/master/twkb.md
+
+const (
+	twkbTypePoint              = 1
```
albertteoh, line 7: nit: These constants could technically be successfully interchanged with one another. That is, usages of `twkbTypeMultiPoint ` could be accidentally but successfully replaced with `twkbMaxDimensions` or `twkbHasIDs` and succeed. Maybe we can consider typing them?

```
type geomKinds int
type metaHeaders int
```

albertteoh (approved):
LGTM. Added a non-blocking suggestion. I might have to defer to @peterstace re: the GIS-specific logic.

geom/twkb.go
DoctorLoki: I've introduced types for the geometry kinds and metadata headers. You're right, I think it's cleaner and more comprehensible this way.

DoctorLoki:
Note to self: TODO:
* Verify every test case using PostGIS or equivalent, for decoding, and whatever the valid subset is for encoding.
* Determine if we also need methods like `(p Point) AsTWKB()` etc.

geom/twkb_write.go
```diff
@@ -0,0 +1,535 @@
...
+
+func (w *TWKBWriter) writeSize() {
+	// Compute where to store the size data.
+	start := 2
+	if w.hasExt {
+		start++ // Ensure extended precision byte remains before the size value.
+	}
+	numBytes := len(w.twkb) - start
+	if numBytes < 0 {
+		panic("attempt to add size value to buffer lacking TWKB header bytes")
```
albertteoh, line 492: Just checking: we're certain this should _never_ happen?

geom/twkb_write.go
```diff
@@ -0,0 +1,535 @@
...
+	var temp []byte
+	temp = append(temp, w.twkb[:start]...)
+	temp = append(temp, buf[:n]...)
+	temp = append(temp, w.twkb[start:]...)
+	w.twkb = temp
+}
+
+func (w *TWKBWriter) writeSize() {
+	// Compute where to store the size data.
+	start := 2
```
albertteoh, line 486: Curious, why 2?

geom/twkb_write.go
```diff
@@ -0,0 +1,535 @@
...
+}
+
+func (w *TWKBWriter) writeIDList(num int) error {
+	if !w.hasIDs {
+		return nil
+	}
+	if num != len(w.idList) {
+		return fmt.Errorf("unexpected ID list length %d, expected %d", len(w.idList), num)
+	}
+	for i := 0; i < num; i++ {
```
albertteoh, line 513: nit: perhaps the more idiomatic approach is `for _, id := range w.idList { w.writeSignedVarint(id) }`?

geom/twkb_write.go
```diff
@@ -0,0 +1,535 @@
...
+
+func (w *TWKBWriter) writeTypeAndPrecision(kind twkbGeometryType) {
+	w.kind = kind
+	w.writeByte(byte(EncodeZigZagInt32(int32(w.precXY))<<4) | byte(w.kind))
+}
+
+func (w *TWKBWriter) writeIsEmptyHeader() {
+	w.isEmpty = true
+	w.writeMetadataHeader(twkbIsEmpty)
+	// Do not write any extended info, bbox, size, or ids.
```
albertteoh, line 371: Just checking: did you mean to place this comment here?

geom/twkb_write.go
```diff
@@ -0,0 +1,535 @@
...
+
+	for i := 0; i < numPolygons; i++ {
+		poly := mp.PolygonN(i)
+		w.writePolygonRings(poly)
+	}
+	return nil
+}
+
+func (w *TWKBWriter) writeGeometryCollection(gc GeometryCollection) error {
+	w.writeTypeAndPrecision(twkbTypeGeometryCollection)
```
albertteoh, line 335: It seems (I could be wrong) that each geom's write method has a common block of code to start with:

```
	w.writeTypeAndPrecision(<the type enum>)

	if ctype := gc.CoordinatesType(); ctype != w.ctype {
		return fmt.Errorf("mismatched coordinate dimensions got %s expected %s", ctype, w.ctype)
	}

	if gc.IsEmpty() {
		w.writeIsEmptyHeader()
		return nil
	}
	w.writeInitialHeaders()
```

Perhaps this could be extracted out into a method?

geom/twkb_write.go
```diff
@@ -0,0 +1,535 @@
...
+		w.ctype = DimXYZ
+	} else if hasM {
+		w.ctype = DimXYM
+	} else {
+		w.ctype = DimXY
+	}
+
+	if hasZ || hasM {
+		w.hasExt = true
+	}
```
albertteoh, line 99:
```suggestion
	w.hasExt = hasZ || hasM
```

geom/twkb_write.go
```diff
@@ -0,0 +1,535 @@
...
+			ival -= w.refpoint[d]
+			n := binary.PutVarint(buf[:], ival)
+
+			w.twkb = append(w.twkb, buf[:n]...)
+			w.refpoint[d] += ival
+			c++
+		}
+		if !w.bboxValid {
+			w.bboxValid = true
+		}
```
albertteoh, line 448: Maybe I'm missing something but why is there a need to check `if !w.bboxValid`?

geom/twkb_parser.go
```diff
@@ -0,0 +1,528 @@
...
+// listed in its header information.
+func UnmarshalTWKBWithHeaders(twkb []byte, opts ...ConstructorOption) (g Geometry, bbox []int64, ids []int64, err error) {
+	p := newTWKBParser(twkb, opts...)
+	g, err = p.nextGeometry()
+	return g, p.bbox, p.idList, p.annotateError(err)
+}
+
+// TWKBParser holds all state information for interpreting TWKB buffers
+// including information such as the last reference point used in coord deltas.
+type TWKBParser struct {
```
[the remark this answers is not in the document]
DoctorLoki: I've made `TWKBParser` and `TWKBWriter` unexported now.

geom/twkb_write.go
DoctorLoki: Yes, it'd be a programming error if we're trying to write additional headers before the initial headers have been written. I've checked all the cases, so this shouldn't ever happen, which is why I've written it as a `panic`.

geom/twkb_write.go
DoctorLoki: That comment at the end of the function was to explain why there's nothing else to be done in that function (because it's an empty object so we stop right there). I'll rewrite that to make it clearer.

geom/twkb_write.go
DoctorLoki: Although it's true there's common(ish) code at the top of each geometry type's write function, the behaviour of `IsEmpty()` for the various types may mean we want to treat them differently. I'll do some more testing before considering amalgamating any of this code.

geom/twkb_write.go
DoctorLoki: In general that transformed code is not identical (but I think it is valid in this case). I'll make the change.

geom/twkb_write.go
DoctorLoki: Yep, you're right, that was a remnant of some refactored code. It was originally inside the inner loop within an `if` statement having the same condition, but that was incorrect because it would set the `bboxValid` flag to true after the X co-ordinate was examined, rather than after the entire first point was examined (which had a weird side-effect of making the bounding box min and max extents wrong for all dimensions except for X).

I'll add a comment to explain why the bbox becomes valid after the first point, i.e. why the code is where it is.

geom/twkb_write.go
DoctorLoki: I've now separated the TWKB writer's internal buffer into three separate buffers, and I think it does clarify things, so I've committed that version.

DoctorLoki:
Current status of this code:
* I have verified that every test case matches how PostGIS functions. It is only the `IsEmpty()` cases that may need to be adjusted. See discussion earlier.
* I haven't yet implemented methods such as `(p Point) AsTWKB()` but it would be straightforward to do so.

peterstace:
Thanks for putting so much work into this! I should get an opportunity to take a deep look either this weekend on early next week 😄

geom/twkb_parser.go
DoctorLoki: Update: I'm fairly convinced that the currently implemented approach is correct. That is, if a polygon ring's final point is the same as its first point, the TWKB decoder silently allow that, but also allow the case that the final point isn't the same as its first point (in which case close the ring by appending the first point as its final point). This seems to be what PostGIS does.

My implementation also allows both forms to be generated from the TWKB encoder, with a Boolean flag to control that. Obviously, for size of output it's generally better to omit the final point of a ring as it is redundant information. This seems not to be what PostGIS does, it seems to always output the final point, and there doesn't seem to be a way to control this behaviour. This simply means my implementation is a little more versatile, which is not a problem.

geom/twkb_parser.go
```diff
@@ -0,0 +1,528 @@
...
+				if first != final {
+					finalPointDiffersFromFirst = true
+					break
+				}
+			}
+
+			if finalPointDiffersFromFirst {
+				// Append first point again, to close the ring.
+				for d := 0; d < p.dimensions; d++ {
+					coords = append(coords, coords[d])
```
peterstace, line 335: Very minor nit: could do `coords = append(coords, coords[:p.dimensions]...)` instead of the loop.

geom/twkb_parser.go
```diff
@@ -0,0 +1,528 @@
...
+	c := 0
+	for i := 0; i < numPoints; i++ {
+		for d := 0; d < p.dimensions; d++ {
+			val, err := p.parseSignedVarint()
+			if err != nil {
+				return nil, fmt.Errorf("point %d of %d coord %d varint malformed: %w", i, numPoints, d, err)
+			}
+
+			p.refpoint[d] += val // Reverse coord differencing to find the true value.
+			coords[c] = float64(p.refpoint[d]) * p.scalings[d]
```
peterstace, line 482: The conversion from the `p.refpoint[d]` value to `float64` is technically lossy above 2^52, since floats start "skipping" integer values above that point. I don't really see an easy way around it though... (e.g. we probably _don't_ want to mess around with arbitrary precision arithmetic). I don't think it will be a problem in practice, but thought I'd bring it up anyway in case you see an easy solution.

peterstace (commented):
I've reviewed the parsing only so far (I'll get to the writer soon!). Everything is looking pretty good so far!

geom/twkb.go
```diff
@@ -0,0 +1,54 @@
...
+
+const (
+	twkbHasBBox    twkbMetadataHeader = 1
+	twkbHasSize    twkbMetadataHeader = 2
+	twkbHasIDs     twkbMetadataHeader = 4
+	twkbHasExtPrec twkbMetadataHeader = 8
+	twkbIsEmpty    twkbMetadataHeader = 16
+)
+
+// DecodeZigZagInt32 accepts a uint32 and reverses the zigzag encoding
```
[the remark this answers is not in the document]
DoctorLoki: Sure, the ZigZag code can be unexported, it was exported really only for testing.

geom/twkb_parser.go
```diff
@@ -0,0 +1,528 @@
...
+func UnmarshalTWKB(twkb []byte, opts ...ConstructorOption) (Geometry, error) {
+	p := newtwkbParser(twkb, opts...)
+	g, err := p.nextGeometry()
+	return g, p.annotateError(err)
+}
+
+// UnmarshalTWKBWithHeaders parses a Tiny Well Known Binary (TWKB),
+// returning the corresponding Geometry, and any bounding box and any IDs
+// listed in its header information.
+func UnmarshalTWKBWithHeaders(twkb []byte, opts ...ConstructorOption) (g Geometry, bbox []int64, ids []int64, err error) {
```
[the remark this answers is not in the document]
DoctorLoki: Your understanding of the bounding box situation is correct in all ways. I'll add the suggested cheap function for just reading the bounding box header if it exists in the data stream.

geom/twkb_write.go
```diff
@@ -0,0 +1,534 @@
+package geom
+
+import (
+	"encoding/binary"
+	"fmt"
+	"math"
+)
+
+// MarshalTWKB accepts a geometry and generates the corresponding TWKB byte slice.
+func MarshalTWKB(geom Geometry,
```
peterstace, line 10: Regarding the function signature, a few things stick out to me:

- The user shouldn't have to supply `hasZ` and `hasM`. Instead, `MarshalTWKB` should be able to get this info using `geom.CoordinatesType().Is3D()` and `geom.CoordinatesType().IsMeasured()`.
- `precXY` should always be specified, but `precZ` and `precM` only make sense for 3D and measured geometries.
- My understanding of the spec is that writers _shouldn't_ close rings. So I don't think we need to make that behaviour configurable. Does that match your understanding?
- `hasSize` and `hasBBox` seem sensible to allow to be configured, although IMO ideally we would default these to both `false`.
- `idList` is sensible to allow, some (many?) users wouldn't need it.

What do you think about using the functional options pattern? The signature could look something like this:

```
func MarshalTWKB(g Geometry, precXY int, opts ...TWKBWriterOption) ([]byte, error) {
    var os twkbWriterOptionSet
    os.precZ = precXY // default if not set by opt
    os.precY = precXY // default if not set by opt
    for _, opt := range opts {
        opt(&os)
    }
    ... // the rest of the logic
}

type TWKBWriterOption func(*twkbWriterOptionSet)

type twkbWriterOptionSet struct {
    precZ, precM int
    hasSize, hasBBox bool
    idList []int64
}

func TWKBPrecisionZ(int precZ) TWKBWriterOption {
    return func(s *twkbWriterOptionSet) {
        s.precZ = precZ
    }
}

... // the rest of the options

```

geom/twkb_write.go
```diff
@@ -0,0 +1,534 @@
...
+		other.precM,      // Same precision as in parent.
+		other.hasSize,    // If parent is using a size header, child should too.
+		false,            // No bbox in sub-geometries.
+		other.closeRings, // If parent requires closed rings, child should too.
+		nil,              // No ID list in sub-geometries.
+	)
+}
+
+func (w *twkbWriter) formTWKB() []byte {
+	var data []byte
```
peterstace, line 133: Nit: since you know the total number of bytes needed (the sum of the header, bbox, and contents), you could pre-allocate the data slice.

peterstace (changes requested):
Overall, looks good! A few things to discuss though.

geom/twkb_parser.go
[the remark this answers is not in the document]
DoctorLoki: Concerning the bounding box header parsing:

I've modified the function signature for
```UnmarshalTWKBWithHeaders(twkb []byte, opts ...ConstructorOption) (g Geometry, bbox []Point, ids []int64, err error)```

Previously that function was returning bounding box information as integers (which exposed implementation internals, naughty!) but now the bounding box is returned as a slice of `Point`. If the slice is empty, there was no bounding box header. Otherwise it contains a minimum and a maximum `Point`. Because `Point` is used, the X, Y, Z and M coordinate values may all be present; no data loss occurs.

I've also added two new functions:
```UnmarshalTWKBBoundingBoxHeader(twkb []byte) (bbox []Point, err error)```
and
```UnmarshalTWKBEnvelope(twkb []byte) (bbox Envelope, hasBBox bool, err error)```

The former function only parses the bounding box header, so it returns without parsing the full geometry.
The semantics of the `Point` slice are the same as discussed above.

The latter function also only parses the bounding box header, but returns it as an `Envelope`.
The `hasBBox` flag is false if there's no such header. Otherwise, the envelope is valid.
Due to the definition of `Envelope` only X and Y values can be reported this way.
(Any Z and M values are omitted because they cannot be represented in an `Envelope`.)

I've added tests checking all these functions behave correctly.

File: geom/twkb_parser.go [hunk 24]
```diff
@@ -0,0 +1,528 @@
...
+
+// twkbParser holds all state information for interpreting TWKB buffers
+// including information such as the last reference point used in coord deltas.
+type twkbParser struct {
+	twkb []byte
+	pos  int
+	opts []ConstructorOption
+
+	kind  twkbGeometryType
+	ctype CoordinatesType
```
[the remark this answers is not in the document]
DoctorLoki: Regarding the information redundancy between `ctype` and `hasZ` and `hasM` and `dimensions` within the TWKB parser:

Although yes there's redundancy of information, the way these values are used differs.
* `hasZ` and `hasM` are derived from the TWKB headers, and thus indicate what was present in the input bytes.
* `dimensions` is then deduced from `hasZ` and `hasM` and controls how coordinates are parsed.
* `ctype` is also deduced from them, and is only used when constructing a `Point` object.

I feel there's a separation here, where `hasZ` and `hasM` and `dimensions` are TWKB-specific decoding state variables, whereas `ctype` is part of the interface for constructing geometries. So to me the separation of those two kinds of variables serves a purpose. The former variables act as documentation which can be read alongside the spec to verify the implementation, the latter `ctype` variable is just a convenience for creating geometries. Shifting all those status flags into the one `ctype` variable I feel might obscure the code a bit.

That being said, I'm happy to combine them if you feel strongly there should be no information redundancy.

geom/twkb_parser.go
DoctorLoki: Regarding converting coordinates to/from `float64`:

There are four bits in the TWKB header to indicate a signed integer for the power of 10 scaling factor for X and Y coordinates. The case we're probably most interested in is where we are keeping 7 decimal places in a latitude or longitude, which is the maximum number of decimal places TWKB lets us retain. Now 10^7 is approximately 2^24, and if we're keeping whole numbers up to +/-180 in front of that, we can assume another 8 bits on top, so it probably amounts to 32 bits of significant information for most real world cases.

My _feeling_ is the `int64` data we're dealing with won't have float conversion roundoffs, and probably won't lose much precision even after scaling. Besides, I don't see a good way around a hypothetical float roundoff, it's simply inherent to the efficiency of using floats.

geom/twkb_write.go
DoctorLoki: Regarding your comments about the encoder:
* Yes, I'll rework the function to allow options. I'd considered it, but figured I'd wait to see what approach you'd prefer.
* Agreed that many users won't use the ID list. It's there for adherence to the spec.
* Agreed that many flags can often be omitted or defaulted to false.
* Technically, supplying `hasZ` and `hasM` separately from the `Geometry` could provide additional functionality of stripping those coordinate values during encoding if they are `false` (although that's not the way the code is currently written). If you're happy to always deduce these from the geometry, those parameters can be omitted as you mentioned.
* You're correct that the spec says the encoder shouldn't close rings. The reason I added that capability was to allow byte-for-byte comparison against an actual PostGIS implementation which does seem to close rings. If possible I'd like to retain the capability, but hide it behind an option.

peterstace: Sounds good!

geom/twkb_parser.go
```diff
@@ -0,0 +1,660 @@
...
+		XY{
+			p.scalings[0] * float64(p.bbox[0]),
+			p.scalings[1] * float64(p.bbox[2]),
+		},
+		XY{
+			p.scalings[0] * float64(p.bbox[0]+p.bbox[1]),
+			p.scalings[1] * float64(p.bbox[2]+p.bbox[3]),
+		},
+	), true, nil
+}
```
peterstace, line 90: The `Envelope` can represent an empty Envelope, so the bool return isn't really needed. Should probably also use the regular envelope constructor rather than the unchecked variant (I don't see any reason not to).

```suggestion
func UnmarshalTWKBEnvelope(twkb []byte) (Envelope, error) {
	p := newtwkbParser(twkb)
	if err := p.parseHeaders(); err != nil {
		return Envelope{}, p.annotateError(err)
	}
	if !p.hasBBox {
		return Envelope{}, nil
	}
	return NewEnvelope([]XY{
		XY{
			p.scalings[0] * float64(p.bbox[0]),
			p.scalings[1] * float64(p.bbox[2]),
		},
		XY{
			p.scalings[0] * float64(p.bbox[0]+p.bbox[1]),
			p.scalings[1] * float64(p.bbox[2]+p.bbox[3]),
		},
	})
}
```

geom/twkb_parser.go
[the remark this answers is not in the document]
peterstace: What you say makes senes, let's leave it as is.

peterstace (commented):
Looking really good! Just a new comment about Envelope being able to be empty, but once that's fixed we should be able to merge 🍰

geom/twkb_parser.go
DoctorLoki: I've modified the envelope function as you suggest, fixing the tests and also adjusting the comment to clarify.

peterstace (approved):

peterstace:
Thank you for contributing this large and significant feature 😁 🍰

PR #444, opened 2022-03-18
Bugfix: `ExactEquals` not correct when both geoms are empty points

peterstace:
## Description

    Fix bug in `ExactEquals` for empty Points
    
    The `ExactEquals` method didn't properly take into consideration
    coordinates type (XY vs. XYZ vs. XYM vs. XYZM) when both Point arguments
    were empty.
    
    The actual fix is only in the `pointsEq` method. In the previous code,
    `ok1` and `ok2` would both evaluate to `false` (because the `Points` are
    both empty). This would then skip the call to `c.eq(c1, c2)`, which was
    where the coordinates type comparison occurred.
    
    The fix is to explicitly check for the case where both `Point`s are
    empty, and handle the coordinates type check there. The `c.eq` method
    still needs to check for coordinate type, to ensure that inputs such as
    `POINT (1 2)` and `POINT Z (1 2 3)` continue to compared as non-equal.

## Check List

Have you:

- Added unit tests? Yes.

## Related Issue

- N/A

## Benchmark Results

N/A

### DIFF
--- a/geom/alg_exact_equals.go
+++ b/geom/alg_exact_equals.go
@@ -155,9 +155,12 @@ func (c exactEqualsComparator) lineStringsEq(ls1, ls2 LineString) bool {
 }
 
 func (c exactEqualsComparator) pointsEq(p1, p2 Point) bool {
+	if p1.IsEmpty() && p2.IsEmpty() {
+		return p1.CoordinatesType() == p2.CoordinatesType()
+	}
 	c1, ok1 := p1.Coordinates()
 	c2, ok2 := p2.Coordinates()
-	return ok1 == ok2 && (!ok1 || c.eq(c1, c2))
+	return ok1 && ok2 && c.eq(c1, c2)
 }
 
 func (c exactEqualsComparator) multiPointsEq(mp1, mp2 MultiPoint) bool {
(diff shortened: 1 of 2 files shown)


geom/alg_exact_equals.go
```diff
@@ -155,9 +155,12 @@ func (c exactEqualsComparator) lineStringsEq(ls1, ls2 LineString) bool {
 }
 
 func (c exactEqualsComparator) pointsEq(p1, p2 Point) bool {
```
peterstace, line 157: This is the actual bugfix.

geom/alg_exact_equals.go
```diff
@@ -155,9 +155,12 @@ func (c exactEqualsComparator) lineStringsEq(ls1, ls2 LineString) bool {
...
+	if p1.IsEmpty() && p2.IsEmpty() {
+		return p1.CoordinatesType() == p2.CoordinatesType()
```
albertteoh, line 159: Can two non-empty points with different coordinate types ever be considered equal? Or perhaps this is handled by `c.eq(c1, c2)`.

peterstace: Yes, that case is handled by the call to `c.eq(c1, c2)`.

albertteoh (approved):

Issue #271, opened 2020-10-07
Revisit Support for GeometryCollection in DCEL algorithm


peterstace:
For the "relates" functions, we currently give an error back if the input is a geometry collection. We should be able to handle geometry collections, _at least_ in the case where they are empty. We may be able to handle them in the more generic cases as well.

peterstace:
This in general applies to all DCEL operations.

PR #445, opened 2022-03-18
GeometryCollection DCEL

davidmcleish:
## Description

Support DCEL built from a `GeometryCollection`. This allows binary operations on `GeometryCollection`s (`Union`, `Intersection` etc.).

Notes:
* There's a bug where if the GeometryCollection contains elements that overlap each other, then some set operations behave badly. Worked around this by taking the union of all the elements of the LHS and RHS first (AFAICT `Union` works) and added a couple of repro tests.
* The output of some set operations on GeometryCollections is nondeterministic. It always produces equivalent geometry but it differs in where it puts intermediate points. (I think the root cause is map iteration order.) Added an `approx` option in `alg_set_op_test.go` to compare the geometry with `Equals` instead of `ExactEquals`.

## Check List

Have you:

- [x] Added unit tests? 
- [ ] Add cmprefimpl tests? (if appropriate?)
- [x] Updated release notes? (if appropriate?)

## Related Issue

https://github.com/peterstace/simplefeatures/issues/271

## Benchmark Results

  `run_benchmarks.sh` script.

<details>

```
PASTE BENCHMARKS HERE
```

</details>


geom/alg_set_op.go
```diff
@@ -13,39 +13,38 @@ func Union(a, b Geometry) (Geometry, error) {
 	if b.IsEmpty() {
 		return a, nil
 	}
-	g, err := setOp(a, b, selectUnion)
+	g, err := setOp(a, b, selectUnion, false)
```
peterstace, line 16: Do you know the reason that union works without first merging members of geometry collections?

geom/alg_set_op.go
```diff
@@ -72,3 +87,15 @@ func setOp(a, b Geometry, include func([2]label) bool) (Geometry, error) {
 	}
 	return g, nil
 }
+
+func merge(gc GeometryCollection) (Geometry, error) {
+	merged := Geometry{}
+	for _, g := range gc.geoms {
```
peterstace, line 93: Just noting that it's valid for a child in a geometry collection to be itself a geometry collection. This nesting can be arbitrarily deep. So in some sense you can think of a geometry collection as a tree with the root being the original geometry collection and any non-leaf nodes below it being additional geometry collections.

The code you've got here should be ok under the assumption that `Union` works for geometry collections.

peterstace (commented):
Thanks for the contribution!

The general approach looks good, although I'm a bit hesitant to progress without first understanding why DCEL operations other than union were causing some issues on certain inputs (and why merging prevents the problems).

I need to spend a bit of time debugging the issue to understand more.

peterstace:
I've got a more minimal reproduction of the problem here: https://github.com/peterstace/simplefeatures/commit/b2aa900052b92fd01d75f0aa072e36935b9f7745

(more minimal from the perspective of printing out a DCEL with a smaller number of edges/vertices/faces)

I'll keep playing around with it and see if I can work out what's going on.

peterstace (approved):
This looks perfect to me! 🍰

Issue #442, opened 2022-03-07
Implements ST_LineInterpolatePoint


radi2015:
Linear Referencing is so useful

```

import (
	"github.com/peterstace/simplefeatures/geom"
)

const DISTMIN = 0.000001

type Segment struct {
	Start, End geom.XY
}

func (s Segment) Length() float64 {
	return s.End.Sub(s.Start).Length()
}

func distValid(dist float64) {
	if dist < DISTMIN {
		panic("dist can not < 0")
	}
}
func (s Segment) Interpolate(dist float64) geom.XY {
	distValid(dist)
	length := s.Length()
	if dist < length {
		rate := dist / length
		return s.Start.Add(s.End.Sub(s.Start).Scale(rate))
	} else {
		return s.End
	}
}

func Interpolate(linestring geom.LineString, dist float64) geom.XY {
	distValid(dist)
	seq := linestring.Coordinates()
	for i := 0; i < seq.Length(); i++ {
		segment := Segment{Start: seq.GetXY(i), End: seq.GetXY(i + 1)}
		length := segment.Length()
		if dist <= length {
			return segment.Interpolate(dist)
		} else {
			dist -= length
		}
	}
	return geom.XY{X: -9999, Y: -9999}
}

```

radi2015:
This is a high-quality space geometry calculation package
I have some immature suggestions
1. LineString with []Point may be easier to understand
2. Coordinates are literally multiple points, and Coordinate may be more appropriate
3. line composed of two points is called segment in JTS. export with Line is also very useful

peterstace:
The reason that `LineString` isn't implemented as `[]Point` is that `Point` may optionally be empty (i.e. WKT `POINT EMPTY`). However, the coordinates that make up a `LineString` cannot be empty.

That's a good suggestion for exposing a `Line` object. There is an internal object that is very similar, I'll take a look to see what the implications of exposing it might be.

peterstace:
This was fixed. by #446 a while ago.

Issue #380, opened 2021-07-17
Add Linear Referencing Functionality


peterstace:
See `5.18. Linear Referencing` in https://postgis.net/docs/reference.html#Geometry_Accessors

PR #446, opened 2022-04-01
Linear Interpolation Methods

peterstace:
## Description

    Add linear referencing methods to LineString
    
    The first method is `InterpolatePoint`, which interpolates a single
    `Point` given a fraction along the `LineString`.
    
    The second method is `InterpolateEvenlySpacedPoints`, which interpolates
    many `Point`s along a `LineString` at once.

## Check List

Have you:

- Added unit tests? Yes.

## Related Issue

- https://github.com/peterstace/simplefeatures/issues/380

- https://github.com/peterstace/simplefeatures/issues/442

## Benchmark Results

N/A

### DIFF
--- a/geom/alg_linear_interpolation.go
+++ b/geom/alg_linear_interpolation.go
@@ -53,5 +53,5 @@ func (l linearInterpolator) interpolate(frac float64) Point {
 }
 
 func lerp(a, b, ratio float64) float64 {
-	return a*(1-ratio) + b*ratio
+	return a + ratio*(b-a)
 }
--- a/geom/alg_linear_interpolation.go
+++ b/geom/alg_linear_interpolation.go
@@ -1,0 +1,57 @@
+package geom
+
+import (
+	"math"
+	"sort"
+)
+
+type linearInterpolator struct {
+	seq        Sequence
+	cumulative []float64
+	total      float64
+}
+
+func newLinearInterpolator(seq Sequence) linearInterpolator {
+	n := seq.Length()
+	if n == 0 {
+		panic("empty seq in newLinearInterpolator")
+	}
+	var total float64
+	cumulative := make([]float64, n-1)
+	for i := 0; i < n-1; i++ {
+		total += seq.GetXY(i).distanceTo(seq.GetXY(i + 1))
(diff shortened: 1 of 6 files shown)


geom/alg_linear_interpolation.go
```diff
@@ -0,0 +1,57 @@
...
+			Y: lerp(p0.Y, p1.Y, partial),
+		},
+		Z:    lerp(p0.Z, p1.Z, partial),
+		M:    lerp(p0.M, p1.M, partial),
+		Type: l.seq.CoordinatesType(),
+	}.asUncheckedPoint()
+}
+
+func lerp(a, b, ratio float64) float64 {
+	return a*(1-ratio) + b*ratio
```
albertteoh, line 56: I feel the following might be easier to understand:
```
a + ratio*(b - a)
```

At least, my mental model is that, given a line (on the X or Y, etc. plane), compute its length and get the ratio of that length, then add it to the start of the line. It's quite possible my understanding/math is incorrect as well. 😄 

What do you think?

albertteoh (approved):
Nice job! 👍🏼

geom/alg_linear_interpolation.go
peterstace: That's good feedback. My mental modal (which might be a bit more complicated) is that we're "mixing" `a` and `b` together using a linear combination that sums to `1.0`. So `a` gets `1-ratio` of the mix, and `b` gets `ratio` of the mix. It's a bit confusing though, because the `1-ratio` goes on the _first_ point, rather than the second.

I think your way of formulating it is a bit easier to follow geometrically, I've changed to formulation in e2a08b77e211207bcd44418dc96abddad9431bbd.

peterstace:
@albertteoh , thanks for the review 😁

geom/alg_linear_interpolation.go
peterstace: I ran into this article, which gives some interesting insight into a better way to write a LERP function: http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p0811r2.html Issue is here: https://github.com/peterstace/simplefeatures/issues/447

Issue #379, opened 2021-07-17
Add DumpRings


peterstace:
Add a new method `DumpRings` to the `Polygon` type.

It's signature should  be `DumpRings ([]LineString)`.

For empty polygons, it should return a zero-length slice. For non-empty polygons, it should contain a slice consisting of the rings in the polygon. The first ring in the slice should be the exterior ring, and the subsequent elements in the slice should be the interior rings (if they exist) in the same order used when creating the polygon.

Be careful not to give the caller of `DumpRings` access to the raw slice held internally by `Polygon`, since they might modify it (and all geometries, including Polygon, should be immutable after creation).

PR #449, opened 2022-05-13
Dump Rings

peterstace:
## Description

    Add `DumpRings` method to `Polygon` type
    
    This gives the rings in the polygon as a slice.

## Check List

Have you:

- Added unit tests? Yes.

## Related Issue

- https://github.com/peterstace/simplefeatures/issues/379

## Benchmark Results

N/A

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@
 - Add initial linear referencing methods to `LineString`. The initial methods
   are `InterpolatePoint` and `InterpolateEvenlySpacedPoints`.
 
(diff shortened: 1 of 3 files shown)


albertteoh (approved):
LGTM. 

The linter errors are questionable to me. e.g.:
- missing return, but there's a panic.
- undeclared name: but it's declared with the `.` notation so that seems okay to me

Curious why they kicked in now and not before. Maybe because the "Unchanged files with check annotations" is a recently introduced "Beta" feature.

If it's blocking this PR, we could consider disabling the linter.

peterstace:
Thanks for reviewing!

Issue #439, opened 2022-01-26
TransformXY doesn't preserve coordinates type for some empty collections


peterstace:
The following program exhibits the buggy behaviour:

```
package main

import (
    "fmt"

    "github.com/peterstace/simplefeatures/geom"
)

func main() {
    tx("POLYGON Z EMPTY")
    tx("MULTILINESTRING Z EMPTY")
    tx("MULTIPOINT Z EMPTY")
    tx("GEOMETRYCOLLECTION Z EMPTY")
}

func must(err error) {
    if err != nil {
        panic(err)
    }
}

func tx(wkt string) {
    g, err := geom.UnmarshalWKT(wkt)
    must(err)
    t, err := g.TransformXY(func(in geom.XY) geom.XY { return in })
    must(err)
    fmt.Println(t.AsText())
}
```

Outputs:
```
POLYGON Z EMPTY
MULTILINESTRING EMPTY
MULTIPOINT EMPTY
GEOMETRYCOLLECTION Z EMPTY
```

The correct output should be:

```
POLYGON Z EMPTY
MULTILINESTRING Z EMPTY
MULTIPOINT Z EMPTY
GEOMETRYCOLLECTION Z EMPTY
```

PR #451, opened 2022-05-13
Bugfix: Coordinates type lost during transform of empty collection

peterstace:
## Description

Fixes a bug in `TransformXY` where the coordinates type would be downgraded to
XY in the case of empty collections.

This occurs because the constructor funcs for MultiPoint and multiLineString
collections infer the coordinate type based on the elements of the collection.
If there are no elements, then there's nothing to infer from, so XY is chosen
as the default. The fix is to explicitly detect this case and manually
construct an empty collection with the correct coordinates type.

## Check List

Have you:

- Added unit tests? Yes.

## Related Issue

- https://github.com/peterstace/simplefeatures/issues/439

## Benchmark Results

N/A


albertteoh (approved):
LGTM

peterstace:
Thanks for the review!

Issue #432, opened 2021-11-16
OmitInvalid should return geometries with the original coordinates Dimension


peterstace:
Consider the following program:

```
func main() {
    seq := geom.NewSequence([]float64{1, 2, 3}, geom.DimXYZ)
    ls, err := geom.NewLineString(seq, geom.OmitInvalid)
    panicIfErr(err)
    fmt.Println(ls)
    fmt.Println(ls.AsText())

    seq = geom.NewSequence([]float64{1, 2, 3, 4, 5, 6}, geom.DimXYZ)
    ls, err = geom.NewLineString(seq, geom.OmitInvalid)
    panicIfErr(err)
    fmt.Println(ls)
    fmt.Println(ls.AsText())
}
```

It emits the following output:

```
LineString[XY] with 0 points
LINESTRING EMPTY
LineString[XYZ] with 2 points
LINESTRING Z (1 2 3,4 5 6)
```

In many ways, this is surprising behaviour. It _should_ instead output the following:

```
LineString[XYZ] with 0 points
LINESTRING Z EMPTY
LineString[XYZ] with 2 points
LINESTRING Z (1 2 3,4 5 6)

```

This applies to not just LineStrings, but other concrete geometry types that deal with the `OmitEmpty` constructor option.

PR #458, opened 2022-07-08
Fix OmitInvalid bug for LineString and Polygon

peterstace:
## Description

The bug occurred when an invalid LineString or Polygon was constructed using the OmitInvalid constructor option. It resulted in the Polygon or LineString correctly being replaced with an EMPTY variant, however the coordinates type was always set to DimXY.

The bugfix sets the coordinates type to match the input in the omit invalid case.

## Check List

Have you:

- Added unit tests? Yes.

## Related Issue

- https://github.com/peterstace/simplefeatures/issues/432

## Benchmark Results

N/A

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## Unreleased
+
+- Fix a bug where the original coordinates type was not retained when using the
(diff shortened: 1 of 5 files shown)


albertteoh (approved):
LGTM

peterstace:
Thanks for the review @albertteoh !! 🍰 😁

Issue #381, opened 2021-07-17
Add Bounding Diagonal 


peterstace:
See https://postgis.net/docs/ST_BoundingDiagonal.html

- Add a new method to `Envelope` with signature `Diagonal() LineString`.
- It should return a LineString from `Min()` to `Max()`.
- It should return an empty `LineString` in the case where the `Envelope` represents a single point.

peterstace:
Returning an empty `LineString` when `Envelope` represents a single point doesn't seem quite right to me. Note that PostGIS returns an _invalid_ line string in that case (the start and end points are the same).

It would be better instead to return a `Geometry`, with it being a `Point` or empty `GeometryCollection` in the degenerate cases, and a `LineString` in the normal case.

PR #489, opened 2023-02-04
Add `BoundingDiagonal()` method to `Envelope`

peterstace:
## Description

The main usecase for this method is to help a user construct a compact representation of an envelope, for example as WKB. Rather than encoding the polygon representation of the Envelope, the LineString representation can be used instead.

The Polygon representation uses 89 bytes (1 for the byte order, 4 to encode the number of rings, 4 to encode the number of points in the outer ring, and then 80 to represent the 5 points in the ring itself).  The LineString representation only uses 37 bytes (1 for the byte order, 4 to encode the number
of points, and 32 to represent 2 points).

## Check List

Have you:

- Added unit tests? Yes.

## Related Issue

- https://github.com/peterstace/simplefeatures/issues/381

## Benchmark Results

N/A, this is new functionality that doesn't have benchmark tests.

### DIFF
--- a/geom/type_envelope.go
+++ b/geom/type_envelope.go
@@ -295,3 +295,27 @@ func (e Envelope) box() (rtree.Box, bool) {
 		MaxY: e.maxY,
 	}, !e.IsEmpty()
 }
+
+// BoundingDiagonal returns the LineString that goes from the minimum point in
+// the Envelope to the point in the Envelope. If the envelope is degenerate and
+// represents a single point, then a Point is returned instead of a LineString.
+// If the Envelope is empty, then the empty Geometry (representing an empty
+// GeometryCollection) is returned.
+func (e Envelope) BoundingDiagonal() Geometry {
+	if e.IsEmpty() {
+		return Geometry{}
+	}
+	if e.IsPoint() {
+		return e.min().asUncheckedPoint().AsGeometry()
+	}
+
+	// The Envelope has already been validated, so it's safe to skip validation
+	// when constructing the diagonal.
+	coords := []float64{e.minX(), e.minY, e.maxX, e.maxY}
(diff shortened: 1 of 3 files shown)


geom/type_envelope.go
```diff
@@ -295,3 +295,27 @@ func (e Envelope) box() (rtree.Box, bool) {
 		MaxY: e.maxY,
 	}, !e.IsEmpty()
 }
+
+// BoundingDiagonal returns the LineString that goes from the minimum point in
+// the Envelope to the point in the Envelope. If the envelope is degenerate and
```
albertteoh, line 300: `from the minimum point in the Envelope to the point in the Envelope`: to which point in the Envelope?

What do you think about describing this as a LineString going from the bottom-left vertex to the top-right vertex of the envelope? At least, that's my understanding of the code, but perhaps my understanding/terminology is too plebeian. 😅

albertteoh (approved):

geom/type_envelope.go
peterstace: Ah, I should have said "from the minimum point in the Envelope to the maximum point in the Envelope".

> What do you think about describing this as a LineString going from the bottom-left vertex to the top-right vertex of the envelope?

That would work if we could assume that geometries were always in a coordinate system where X increases from left to right and Y as you move _up_ the map. WGS84 and the UTM projections would be one such example. Unfortunately it doesn't hold for other projections, such as Web Mertcator, which has Y increasing as you move _down_ the map.

I've reworded the comment in 30468163541b696d898ce5ca5a93e8eb9f249eb3 to explicitly refer to the `Min` and `Max` methods, which is uses to construct the `LineString`.

peterstace:
Thanks for reviewing! 🙇

PR #491, opened 2023-02-14
Simplify the `Type` method of the `Geometry` type

peterstace:
## Description

No need for the switch statement since we already have direct access to the geometry type. The switch statement is likely a remnant of extracting out geometry type (a _long_ time ago). The only functional difference here is that if a Geometry is somehow constructed with an invalid geometry type (which should be impossible), then the `Type` method will no longer panic but instead just return the invalid geometry type.

## Check List

Have you:

- Added unit tests? Not needed, since there's no testable functional change.

- Updated release notes? (if appropriate?) N/A, not a change that's visible to users.

## Related Issue

- N/A

## Benchmark Results

- Skipped, since this definitely won't have a performance impact.

### DIFF
--- a/geom/type_geometry.go
+++ b/geom/type_geometry.go
@@ -54,26 +54,9 @@ func (t GeometryType) String() string {
 	}
 }
 
-// Type returns a string representation of the geometry's type.
+// Type returns the type of the Geometry.
 func (g Geometry) Type() GeometryType {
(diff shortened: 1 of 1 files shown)


albertteoh (approved):
👍🏼

peterstace:
Thanks for the review! 😁

Issue #474, opened 2022-11-11
Don't use trigonometry when sorting half edge records radially


peterstace:
In `doublyConnectedEdgeList`'s `fixVertex` method, trigonometry is used to sort edges radially. Specifically, we use the `math.Atan2` function to calculate the angle of each edge, and then sort the edges via their angle.

Some alternate approaches are described here: https://stackoverflow.com/questions/6989100/sort-points-in-clockwise-order. These approaches may be better because they don't rely on trigonometry. Using regular multiplication/addition may be more numerically stable for angles that are extremely close together.

PR #493, opened 2023-02-17
Minor cleanup for radial sorting

peterstace:
## Description

- The radial sort 'less' function is extracted out into its own standalone function, making the parent function (`fixVertex`) not need to know about the details of the sort.

- A new `XY` method `lengthSq` is added, that computes the square of the length of the distance between the origin and the XY value.

- The `Cross` and `lengthSq` methods are used for radial sorting rather than reimplementing the logic for those computations.

## Check List

Have you:

## Related Issue

- https://github.com/peterstace/simplefeatures/issues/474

## Benchmark Results

Not run yet.


albertteoh (approved):

peterstace:
Thanks for reviewing 😁

Issue #482, opened 2022-11-15
Investigate proper GeoJSON winding order


peterstace:
Right now, no normalisation takes place when marshalling GeoJSON. Should ring orientation be normalised? Needs a bit of investigation and thought about what implications this could have.

akhenakh:
In the fully-standardized revised GeoJSON specification (aka [RFC 7946](https://datatracker.ietf.org/doc/html/rfc7946)), polygons should follow the right-hand rule - exterior rings are counterclockwise, interior clockwise. However, for backward-compatibility, this is not required. Polygons with differently-wound coordinates should still work fine.

PR #495, opened 2023-03-20
Orient Polygon and MultiPolygon rings before GeoJSON marshalling

peterstace:
## Description

This is to better follow RFC7946. See https://datatracker.ietf.org/doc/html/rfc7946 and linked issue for details.

## Check List

Have you:

- Added unit tests? Yes.

## Related Issue

- https://github.com/peterstace/simplefeatures/issues/482

### DIFF
--- a/geom/geojson_marshal_test.go
+++ b/geom/geojson_marshal_test.go
@@ -1,7 +1,9 @@
 package geom_test
 
 import (
+	"encoding/hex"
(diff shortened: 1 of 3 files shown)


peterstace:
I'm getting some reference implementation check failures against PostGIS. I think that's to be expected since PostGIS doesn't do ring normalisation.

Issue #496, opened 2023-03-23
panic with Union


akhenakh:
Union of those 2 geometries panic.

```
panic: runtime error: slice bounds out of range [1:0]
```

Both geometries are reported valid by GEOS, not sure what happened 

play https://go.dev/play/p/U4ZuTk678WI

```go
package main

import (
	"encoding/json"
	"fmt"
	"log"
	"strings"

	"github.com/peterstace/simplefeatures/geom"
)

func main() {
[...]

Issue #486, opened 2023-01-10
Store `cmprefimpl` strings in a separate file


peterstace:
Currently, the input strings for the `cmprefimpl` tests are scraped from the `simplefeatures` codebase. The intention is that the unit tests in the codebase contain a lot of WKT literals, which are perfect inputs for the `cmprefimpl` tests.

This approach has the downside that the addition of new unit tests can sometimes cause the `cmprefimpl` tests to fail. Any additional unit tests (with novel inputs) can create additional `cmprefimpl` test cases, which may fail.

It would be nice if the `cmprefimpl` inputs were decoupled from the unit tests. One way to do this could be to:

- Store the inputs to `cmprefimpl` in a text file (under version control).
- Create a utility to scrape the codebase to populate the text file. This utility would be run manually.
- The utility might need blacklisting functionality.

PR #497, opened 2023-03-27
Give error rather than panic when no rings to extract

peterstace:
## Description

Sometimes (due to numeric precision issues) some internal assumptions during DCEL ring extraction are violated. This change causes those violations to result in an error being returned from functions such as Union and Intersection. The previous (implicit) behaviour was to panic.

This is a stop gap measure put in place until a proper solution can be devised.

## Check List

Have you:

- Added unit tests? Yes.

## Related Issue

- https://github.com/peterstace/simplefeatures/issues/496

## Benchmark Results

N/A


geom/dcel_extract_geometry.go
```diff
@@ -94,6 +94,10 @@ func (d *doublyConnectedEdgeList) extractPolygons(include func([2]bool) bool) ([
 			})
 		}
 
+		if len(rings) == 0 {
```
peterstace, line 97: This is the actual fix for this PR.

internal/cmprefimpl/cmpgeos/checks.go
```diff
@@ -702,6 +702,11 @@ var skipIntersection = map[string]bool{
 	"POLYGON((1 0,-0.9 -0.2,-1 -0.0000000000000032310891488651735,-0.9 0.2,1 0))": true,
 	"LINESTRING(0.5 0,0.5000000000000001 0.5)":                                    true,
 	"LINESTRING(0.5 1,0.5000000000000001 0.5)":                                    true,
+
```
peterstace, line 705: Because of the way that the reference implementation checks work, new test cases get automatically created for any new geometries that are introduced in unit tests. This is nice because we get a lot of test coverage easily, but because the new geometries introduced in unit tests in this PR cause errors for DECL operations, they need to be added to the skip lists here (otherwise they cause the reference implementation checks to fail).

I don't think this is a very sustainable strategy going forward (there are plans to address this in https://github.com/peterstace/simplefeatures/issues/486).

internal/cmprefimpl/cmpgeos/checks.go
```diff
@@ -817,6 +834,11 @@ func checkDCELOp(
 		return nil
 	}
 
+	got, err := op(g1, g2)
```
peterstace, line 837: Previously the skip list was only checked after the simplefeatures operation was executed. This was fine when the skip list was just for skipping comparison of the results. However, now it's also skipping geometries that cause the operation to fail outright. So the skip check has to go before the actual operation is executed.

albertteoh (approved):
LGTM

peterstace:
Thanks for reviewing! 🙇

Issue #364, opened 2021-06-06
Unit tests fail on aarch64


peterstace:
The only test failing are in the `geos` package, in the `TestBuffer` test suite. E.g.

```
--- FAIL: TestBuffer/11 (0.00s)
        libgeos_test.go:716: WKT: LINESTRING(0 0,1 0,1 1)
        libgeos_test.go:719:
            got:  POLYGON((0.9 0.1,0.9 1,0.901921471959677 1.019509032201613,0.9076120467488714 1.0382683432365092,0.9168530387697456 1.0555570233019602,0.9292893218813453 1.0707106781186548,0.9444429766980398 1.0831469612302547,0.961731656763491 1.0923879532511287,0.9804909677983872 1.0980785280403231,1 1.1,1.0195090322016127 1.0980785280403231,1.038268343236509 1.0923879532511287,1.0555570233019602 1.0831469612302547,1.0707106781186546 1.0707106781186548,1.0831469612302544 1.0555570233019602,1.0923879532511287 1.0382683432365092,1.098078528040323 1.019509032201613,1.1 1,1.1 0,1 -0.1,0 -0.1,-0.019509032201612955 -0.09807852804032302,-0.03826834323650912 -0.09238795325112863,-0.055557023301960363 -0.08314696123025443,-0.07071067811865482 -0.07071067811865468,-0.08314696123025457 -0.055557023301960155,-0.0923879532511287 -0.03826834323650893,-0.09807852804032306 -0.01950903220161279,-0.1 0.000000000000000012246467991473533,-0.09807852804032305 0.01950903220161282,-0.0923879532511287 0.03826834323650895,-0.08314696123025456 0.055557023301960176,-0.07071067811865482 0.0707106781186547,-0.05555702330196031 0.08314696123025447,-0.0382683432365091 0.09238795325112864,-0.019509032201612972 0.09807852804032302,0 0.1,0.9 0.1))
[...]

PR #498, opened 2023-03-28
Fixes for `aarch64` CI

peterstace:
## Description

These fixes will allow CI to execute for `aarch64` architecture:

- Use a multi-arch image for `postgis`.

- Add XY tolerance when comparing buffer outputs in GEOS reference implementation tests (because `aarch64` and `x86_64` provide slightly different outputs).

- Adds project names for each docker-compose step, to prevent warnings about orphaned containers (not strictly required for `aarch64`).

## Check List

Have you:

- Added unit tests? N/A

## Related Issue

- https://github.com/peterstace/simplefeatures/issues/364

### DIFF
--- a/internal/cmprefimpl/cmppg/pg.go
+++ b/internal/cmprefimpl/cmppg/pg.go
@@ -48,6 +48,11 @@ const (
 
 // Unary runs a batch of unary operations on a geometry.
 func (p BatchPostGIS) Unary(g geom.Geometry) (UnaryResult, error) {
+	// Retry several times because requests to PostGIS can intermittently fail.
+	// Requests to PostGIS could fail for many reasons. For example, a previous
+	// test case could cause PostGIS to segfault, causing it to restart (so it
+	// may not yet be ready to accept connections by the time this test
+	// executes).
 	var i int
 	for {
 		u, err := p.tryUnary(g)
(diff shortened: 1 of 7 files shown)


peterstace:
There are still some failures in CI, which I believe are due to the postgres image changes. I'll leave this PR in draft until I've got those sorted out.

peterstace:
CI is now passing again, so this is back for review.

internal/cmprefimpl/cmppg/pg.go
```diff
@@ -46,6 +48,23 @@ const (
 
 // Unary runs a batch of unary operations on a geometry.
 func (p BatchPostGIS) Unary(g geom.Geometry) (UnaryResult, error) {
+	var i int
```
peterstace, line 51: I noticed a little bit of intermittent instability with the new postgis image -- adding some retry seems to fix the issue.

albertteoh: > I noticed a little bit of intermittent instability with the new postgis image -- adding some retry seems to fix the issue.

What do you think about adding ☝🏼 somewhere here? As a reader, I would be wondering what was the motivation for adding this retry logic (and could be hard to find by searching through PRs).

peterstace: Good call, added in b45949abf5f21c5a0284b741d8b4db651a7256f2.

albertteoh (approved):

peterstace:
Thanks for reviewing, greatly appreciated 😁 🍰

PR #500, opened 2023-03-31
Trigger Github Actions only once per commit

peterstace:
## Description

Previously, CI via Github Actions would be triggered multiple times per push if there is an open PR for the corresponding branch. When a developer pushes a new commit, the first run would be triggered by the `push` event, and the second run triggered by the `pull_request` event (because pushing new commits _alters_ the pull request).

## Check List

Have you:

- Added unit tests? N/A

## Related Issue

- N/A

## Benchmark Results

- N/A

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,8 +1,6 @@
 name: build
 on:
   push:
-  pull_request:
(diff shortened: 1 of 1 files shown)


albertteoh (approved):
👍🏼

PR #501, opened 2023-04-01
Fix incorrect line/line intersection on `aarch64`

peterstace:
## Description

Fixes a bug where the result of intersecting two lines segments is sometimes incorrect.

The bug only occurs in `aarch64` and is due to an unanticipated [fused multiply-add](https://en.wikipedia.org/wiki/Multiply%E2%80%93accumulate_operation) (FMA) operation. The FMA causes the cross product method to return very small but non-zero results for linearly dependent inputs (i.e. two vectors that are scaled versions of each other).

The bug does **not** occur on `x64_64`.

## Check List

Have you:

- Added unit tests? Yes.

## Related Issue

- https://github.com/peterstace/simplefeatures/issues/364

### DIFF
--- a/geom/xy_test.go
+++ b/geom/xy_test.go
@@ -66,7 +66,7 @@ func TestXYCross(t *testing.T) {
 		{u: xy(1, 1), v: xy(0, 0), want: 0},
 		{u: xy(0.1, 0.1), v: xy(0, 0), want: 0},
 
-		// Not linearly independent:
+		// Linearly dependent:
 		{u: xy(1, 1), v: xy(2, 2), want: 0},
 		{u: xy(-1, -1), v: xy(2, 2), want: 0},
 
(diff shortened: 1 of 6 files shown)


geom/xy_test.go
```diff
@@ -52,3 +53,40 @@ func TestXYUnit(t *testing.T) {
...
+	}
+	for i, tc := range []struct {
+		u, v geom.XY
+		want float64
+	}{
+		// Contains zero:
+		{u: xy(1, 1), v: xy(0, 0), want: 0},
+		{u: xy(0.1, 0.1), v: xy(0, 0), want: 0},
+
+		// Not linearly independent:
```
albertteoh, line 69: Maybe could reword this as `Linearly dependent:`?

albertteoh (approved):
Nice discovery!

geom/xy_test.go
peterstace: Good suggestion, altered in 51c7c7327f57a19bf873a4133046c5c16e22e8d5.

peterstace:
Thanks for reviewing!

Issue #502, opened 2023-04-03
Use latest version of GEOS for testing


peterstace:
The version of GEOS currently being used is a bit out of date (v3.7.1).

At least for the reference implementation tests, we should use the latest version of GEOS to reduce the number of hacks needed to align the outputs of GEOS and simplefeatures.

It would be also nice to run the `github.com/peterstace/simplefeatures/geos` package tests on a few different versions of GEOS to make sure it will always compile.

PR #503, opened 2023-05-01
Use GEOS 3.11 for reference implementation tests

peterstace:
## Description

- Upgrades to GEOS 3.11 for the reference implementation checks.

- The hope was that this would reduce the number of "hacks" needed to get the simplefeatures vs. GEOS results to align. However, it actually ended up with more "hacks" being required.

- Being an GEOS 3.11 however does allow for CI to work locally with `aarch64`, because there are less differences between `x86_64` and `aarch64` or GEOS 3.11 vs. GEOS 3.9.

## Check List

Have you:

- Added unit tests? N/A.

- Add cmprefimpl tests? (if appropriate?) Yes.

- Updated release notes? (if appropriate?) Not yet.

## Related Issue

- https://github.com/peterstace/simplefeatures/issues/502


geos/entrypoints_test.go
peterstace, line 1: The changes in this file aren't semantically important, and represent slight differences in behaviour between GEOS 3.7 and GEOS 3.11.

internal/cmprefimpl/cmpgeos/checks.go
peterstace, line 1: Changes in this file are hacks/skips to get things passing with GEOS 3.11.

In general, the reference implementation checks need a bit of a re-think/re-write, so I'm not too concerned about making them a bit messier in the short term.

PR #507, opened 2023-05-02
Add CoverageUnion wrapper to `geos` package

peterstace:
## Description

- Add CoverageUnion wrapper to `geos` package. This is useful for when you want to union together geometries that are noded together and don't overlap because it's much faster than regular union.

## Check List

Have you:

- Added unit tests? Yes.

- Updated release notes? (if appropriate?): Yes.

## Related Issue

- https://github.com/peterstace/simplefeatures/issues/505

### DIFF
--- a/geos/entrypoints.go
+++ b/geos/entrypoints.go
@@ -15,6 +15,16 @@ package geos
 GEOSGeometry *GEOSMakeValid_r(GEOSContextHandle_t handle, const GEOSGeometry* g) { return NULL; }
 #endif
 
+#define COVERAGE_UNION_MIN_VERSION "3.8.0"
+#define COVERAGE_UNION_MISSING ( \
+	GEOS_VERSION_MAJOR < 3 || \
+	(GEOS_VERSION_MAJOR == 3 && GEOS_VERSION_MINOR < 8) \
+)
+#if COVERAGE_UNION_MISSING
+// This stub implementation always fails:
+GEOSGeometry *GEOSCoverageUnion_r(GEOSContextHandle_t handle, const GEOSGeometry* g) { return NULL; }
+#endif
+
 */
 import "C"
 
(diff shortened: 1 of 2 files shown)


geos/entrypoints_test.go
```diff
@@ -815,3 +815,41 @@ func TestMakeValid(t *testing.T) {
...
+		},
+		{
+			// Input constraint violated: inputs overlap.
+			input:   `GEOMETRYCOLLECTION(POLYGON((0 0,0 1,1 0,0 0)),POLYGON((0 0,0 1,1 1,0 0)))`,
+			wantErr: true,
+		},
+		{
+			// Input constraint violated: not noded correctly.
+			input:   `GEOMETRYCOLLECTION(POLYGON((0 0,0 1,1 0,0 0)),POLYGON((0 0,2 0,0 -2,0 0)))`,
+			wantErr: true,
```
peterstace, line 837: Still need to confirm that the errors coming back contain reasonable content.

peterstace:
I noticed a case where the error reporting isn't quite correct. It's due to what I believe may be a bug in GEOS itself (see https://github.com/libgeos/geos/issues/889 for details). I don't think it makes sense to work around the issue in the wrapper layer.

geos/entrypoints.go
```diff
@@ -337,3 +347,29 @@ func MakeValid(g geom.Geometry, opts ...geom.ConstructorOption) (geom.Geometry,
...
+
+// The CoverageUnion function is used to union polygonal inputs that form a
+// coverage, which are typically provided as a GeometryCollection. This method
+// is much faster than other unioning methods, but there are some constraints
+// that must be met by the inputs to form a valid polygonal coverage. These
+// constraints are:
+//
+//  1. all input geometries must be polygonal,
+//  2. the inputs must not overlap, and
+//  3. the input boundaries must be noded in the same way.
```
albertteoh, line 359: Does this sentence mean that the input boundaries share a common edge? Or is there a different meaning?

geos/entrypoints_test.go
```diff
@@ -815,3 +815,63 @@ func TestMakeValid(t *testing.T) {
...
+		},
+		{
+			// Input constraint violated: not noded correctly.
+			input: `GEOMETRYCOLLECTION(
+				POLYGON((0 0,0 1,1 1,1 0,0 0)),
+				POLYGON((0 1,2 1,2 2,0 2,0 1))
+			)`,
+			wantErr: true,
+		},
+		{
```
albertteoh, line 856: Would it be useful adding another test case where only the vertex is shared?

```
		{
			// Input constraint violated: not noded correctly - shared vertex but not edge.
			input: `GEOMETRYCOLLECTION(
				POLYGON((0 0,0 1,1 1,1 0,0 0)),
				POLYGON((0 1,2 1,2 2,1 2,0 1))
			)`,
			wantErr: true,
		},
```

I was also thinking of an example where polygons have no shared nodes (I hope that's the right terminology!); put simply, they're not touching. Although, maybe the tests already somewhat imply this.

peterstace: It's actually a valid coverage if just a vertex is shared (which indicates my doc comment is probably not good enough!). I've added some more test cases in dccf76d65b282d8d2102e0b534bbd6daf550281e as well.

albertteoh (approved):
Looks good!

peterstace:
Thanks for reviewing!

PR #512, opened 2023-06-02
Use Makefile instead of bash script for CI

peterstace:
## Description

Switches to use a Makefile as a "task runner" rather than a script for running CI related tasks. This is useful because developers of simplefeatures often have to run various tasks as part of development (e.g. run some of the test suits or lint code). It is a lot easier to invoke the Makefile (e.g. `make lint`) rather than opening the CI script and coping/modifying out the part that needs to be run.

## Check List

Have you:

- Added unit tests? N/A

## Related Issue

- https://github.com/peterstace/simplefeatures/issues/490


Makefile
```diff
@@ -0,0 +1,61 @@
+DC_RUN = \
+	docker-compose \
```
albertteoh, line 2: I think `docker-compose` is effectively [deprecated](https://www.docker.com/blog/new-docker-compose-v2-and-v1-deprecation/). What do you think about using `docker compose` instead?

albertteoh (approved):
Love it!

Makefile
peterstace: Ah, yes thanks for reminding me. Fixed in 010e2966224da69d69416a790788cebe14d9dcf4.

Makefile
```diff
@@ -0,0 +1,61 @@
+DC_RUN = \
```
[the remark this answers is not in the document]
peterstace: Yep, that's what I was going for -- you save a line for each variable because you can't define Makefile variables on the same line as something else. The downside is that you need to use `$$` rather than `$` due to escaping, but that's not too bad.

peterstace:
Thanks for reviewing! 😀

PR #519, opened 2023-06-15
Add a `String() string` method to `Envelope`

peterstace:
## Description

This is mainly so that users see a useful representation of `Envelope` when they try to print it out. Otherwise they see something quite ugly that reveals the internal implementation of the `Envelope` (e.g. the "XOR NaN" optimisation).

## Check List

Have you:

- Added unit tests? Yes.

## Related Issue

- https://github.com/peterstace/simplefeatures/issues/517

### DIFF
--- a/geom/type_envelope_test.go
+++ b/geom/type_envelope_test.go
@@ -609,3 +609,19 @@ func TestBoundingDiagonal(t *testing.T) {
 		})
 	}
 }
+
+func TestEnvelopeString(t *testing.T) {
+	for i, tc := range []struct {
+		env  Envelope
+		want string
+	}{
+		{Envelope{}, "ENV EMPTY"},
+		{twoPtEnv(1.5, 2.5, 3.5, 4.5), "ENV(1.5 2.5,3.5 4.5)"},
+	} {
+		t.Run(strconv.Itoa(i), func(t *testing.T) {
+			got := fmt.Sprintf("%v", tc.env)
+			t.Log(got)
+			expectTrue(t, got == tc.want)
+		})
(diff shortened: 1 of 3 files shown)


geom/type_envelope_test.go
```diff
@@ -609,3 +609,19 @@ func TestBoundingDiagonal(t *testing.T) {
...
 	}
 }
+
+func TestEnvelopeString(t *testing.T) {
+	for i, tc := range []struct {
+		env  Envelope
+		want string
+	}{
+		{Envelope{}, "ENV EMPTY"},
+		{twoPtEnv(1.5, 2.5, 3.5, 4.5), "ENV(1.5 2.5,3.5 4.5)"},
```
albertteoh, line 619: Was there a requirement to have a shortened version of the work "envelope"? 

I feel ENV is a common shorthand for "environment", and could be a little confusing. Moreover, WKTs use the full word such as POLYGON or MULTIPOLYGON, and it looks like we're using a similar syntax to WKTs, so using ENVELOPE maintains consistency with how WKTs are named.

What do you think?

albertteoh (approved):
Looks good! Just a question around naming.

geom/type_envelope_test.go
peterstace: I hadn't considered that `ENV` could be confused with environment, and I agreed that it definitely could be!

I've updated to `ENVELOPE` in 888d5c6f1427e2b8d3eb86409245b230d98af152.

peterstace:
Thanks for reviewing!

Issue #504, opened 2023-05-02
Consider removing constructor options


peterstace:
Right now, there are only 2 constructor options:

- Disable validations.
- Omit invalid.

The construction options are provided via the functional options pattern via the geometry constructors.

Omit invalid was only really added to "prove" that there was more than one option and therefore the functional options pattern was warranted.

We should consider to:

- Retire the omit invalid construction option.
- Change from functional options to something simpler. E.g. 2 functions for each constructor, `NewPoint` and `NewPointWithoutValidation`.

peterstace:
Another nice thing about having separate "unvalidated" constructors is that those function signatures don't need an error in the return.

peterstace:
See https://github.com/peterstace/simplefeatures/discussions/525 for discussion about this.

PR #520, opened 2023-06-22
Remove `OmitInvalid` constructor option

peterstace:
## Description

This constructor isn't used in any non-trivial way by simplefeatures
itself, and isn't used by any known users.

It was introduced artificially to "show" that there are multiple uses
cases for constructor options (and thus the function options pattern was
warranted). However, the only truly used construction option is
`DisableAllValidations`.

The removal of `OmitInvalid` is the first step to change how geometries
are constructed without validation.

## Check List

Have you:

- Added unit tests? N/A

## Related Issue

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## Unreleased
 
+- **Breaking change**: Removes the `OmitInvalid` constructor option. Users of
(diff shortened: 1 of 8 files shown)


albertteoh (approved):
👍🏼

peterstace:
Thanks for reviewing!

PR #521, opened 2023-07-21
Add Validate method

peterstace:
## Description

Adds a `Validate` method that checks if a geometry is valid after construction (in case validations were disabled).

## Check List

Have you:

- Added unit tests? Yes.

- Updated release notes? (if appropriate?) Not yet.

## Related Issue

- https://github.com/peterstace/simplefeatures/issues/434

## Benchmark Results

  `run_benchmarks.sh` script.

TODO

<details>

```
PASTE BENCHMARKS HERE
```

</details>

### DIFF
--- a/geom/type_geometry.go
+++ b/geom/type_geometry.go
@@ -957,3 +957,24 @@ func (g Geometry) Simplify(threshold float64, opts ...ConstructorOption) (Geomet
 		panic("unknown type: " + g.Type().String())
 	}
 }
+
+func (g Geometry) Validate() error {
+	switch g.gtype {
+	case TypeGeometryCollection:
+		return g.MustAsGeometryCollection().Validate()
+	case TypePoint:
+		return g.MustAsPoint().Validate()
+	case TypeLineString:
+		return g.MustAsLineString().Validate()
+	case TypePolygon:
+		return g.MustAsPolygon().Validate()
+	case TypeMultiPoint:
+		return g.MustAsMultiPoint().Validate()
+	case TypeMultiLineString:
+		return g.MustAsMultiLineString().Validate()
+	case TypeMultiPolygon:
+		return g.MustAsMultiPolygon().Validate()
+	default:
+		panic("unknown type: " + g.Type().String())
+	}
+}
--- a/geom/type_geometry.go
+++ b/geom/type_geometry.go
@@ -958,6 +958,8 @@ func (g Geometry) Simplify(threshold float64, opts ...ConstructorOption) (Geomet
 	}
 }
 
+// Validate checks if the Geometry is valid. See the documentation for each
+// concrete geometry type for details.
 func (g Geometry) Validate() error {
 	switch g.gtype {
 	case TypeGeometryCollection:
--- a/geom/type_geometry.go
+++ b/geom/type_geometry.go
@@ -959,7 +959,7 @@ func (g Geometry) Simplify(threshold float64, opts ...ConstructorOption) (Geomet
 }
 
 // Validate checks if the Geometry is valid. See the documentation for each
-// concrete geometry type for details about the validation rules.
+// concrete geometry's Validate method for details about the validation rules.
 func (g Geometry) Validate() error {
 	switch g.gtype {
 	case TypeGeometryCollection:
--- a/geom/type_geometry.go
+++ b/geom/type_geometry.go
@@ -959,7 +959,7 @@ func (g Geometry) Simplify(threshold float64, opts ...ConstructorOption) (Geomet
 }
 
 // Validate checks if the Geometry is valid. See the documentation for each
-// concrete geometry type for details.
+// concrete geometry type for details about the validation rules.
 func (g Geometry) Validate() error {
 	switch g.gtype {
 	case TypeGeometryCollection:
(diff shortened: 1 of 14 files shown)


geom/type_geometry.go
```diff
@@ -957,3 +957,24 @@ func (g Geometry) Simplify(threshold float64, opts ...ConstructorOption) (Geomet
 		panic("unknown type: " + g.Type().String())
 	}
 }
+
```
peterstace, line 960: - [ ] TODO add doc comment

geom/type_polygon.go
```diff
@@ -33,50 +22,85 @@ type Polygon struct {
...
+		if r.IsEmpty() {
+			return defyRingNotEmpty.err()
+		}
 		if !r.IsClosed() {
-			return validationError{"polygon ring not closed"}
+			return defyRingClosed.errAtXY(r.Coordinates().GetXY(0))
 		}
 		if !r.IsSimple() {
-			return validationError{"polygon ring not simple"}
+			return defyRingSimple.errAtXY(r.Coordinates().GetXY(0))
```
peterstace, line 91: - [ ] TODO: use the point where the polygon is actually not simple, rather than just an arbitrary point

geom/type_multi_polygon.go
```diff
@@ -9,59 +9,71 @@ import (
...
 type MultiPolygon struct {
 	// Invariant: ctype matches the coordinates type of each polygon.
 	polys []Polygon
 	ctype CoordinatesType
 }
 
-// NewMultiPolygon creates a MultiPolygon from its constituent Polygons. It
-// gives an error if any of the MultiPolygon assertions are not maintained. The
+// NewMultiPolygon creates a MultiPolygon from its constituent Polygons. The
 // coordinates type of the MultiPolygon is the lowest common coordinates type
```
albertteoh, line 21: I'm having a bit of trouble parsing this sentence, especially the "type its Polygons" part: 

> The coordinates type of the MultiPolygon is the lowest common coordinates type its Polygons.

geom/type_multi_polygon.go
```diff
@@ -91,8 +103,8 @@ func validateMultiPolygon(polys []Polygon, opts ctorOptionSet) error {
 				polyBoundaries[j],
 			)
 			if !interMLS.IsEmpty() {
-				return validationError{"multipolygon child polygon " +
-					"boundaries intersect at multiple points"}
+				return defyChildPolysMultiTouch.errAtPt(
```
albertteoh, line 106: What does `defy` mean in the context of reporting errors here?

geom/type_polygon.go
```diff
@@ -33,50 +22,85 @@ type Polygon struct {
 // any inner rings follow. If no rings are provided, then the returned Polygon
 // is the empty Polygon. The coordinate type of the polygon is the lowest
 // common coordinate type of its rings.
+//
+// An error is returned if any of the Polygon constraints are not met (see the
+// Validate method for details). Note that the validity of each input
+// LineString is not verified.
```
albertteoh, line 28: I noticed validity of input geometries is not checked for certain geometries, esp. Multi* and here. Is this because validation is too computationally expensive to execute as default behaviour in a ctor?

albertteoh (approved):

geom/type_multi_polygon.go
peterstace: Thanks, yep I was missing an "of". Fixed in 8425f76b10062fe1fe5a62ede9998952b435a387.

geom/type_multi_polygon.go
peterstace: `defy` refers to the validation rule being defied. I think there are two issues here, though:

- `defyRingsMultiTouch` and `defyChildPolysMultiTouch` as names are inverted compared to the others (i.e. they describe what's wrong rather than what the rule is).
- `defy` is short, but maybe not the right word to use. `violation` is probably better.

I'll have a think about better ways to name these.

geom/type_polygon.go
peterstace: Sort of. We definitely don't want to validate them multiple times, because they're not cheap to run (especially the Polygon validation).

The idea is that Multi*, and GeometryCollection are all collection-style geometries, and provide additional constraints between elements of those collections. An argument could be made that Polygon should revalidate its input rings (which is cheap anyway) because it's _not_ a collection.

I'll have a bit more of a think about this.

geom/type_multi_polygon.go
albertteoh: Yup, `violation` sounds good. Other alternatives are `break` or `breach`. I thought [this is a nice explanation](https://www.italki.com/en/post/question-231972) of these variants:

> 1) "breach" usually refers to a barrier of some sort, such as a wall. e.g. "Sir, the wall has been breached by the enemy. Should we fall back now??"
> 2) "break" is the most commonly used of the above four words, and can be used in almost any context
> 3) "violate" usually refers to a law or right, such as "that country violates it's peoples' human rights" or a person's integrity ("violate" in that context is often synonymous with rape")
> 4) "defy" usually refers to a regime, or some kind of other person or institution, it usually infers willful action against another person's wishes

peterstace: Cool thanks for that explanation. I think those descriptions confirm that violate is the best word to use. Updated in da510a0b49fd2a9d522148abf06462255c7baccf.

geom/type_polygon.go
peterstace: > I'll have a bit more of a think about this.

FYI I went with making the polygon constructor revalidate its rings (since polygons are categorically different than multi* types).

PR #526, opened 2023-09-15
Don't validate results from GEOS

peterstace:
## Description

Recent versions of GEOS use OverlayNG (NG stands for New Generation) for set operations between geometries (Intersection, Union, etc.). With the introduction of OverlayNG, invalid geometries are no longer frequently produced by GEOS. They are either non-existent, or very rare.

Before this change, the `simplefeatures` GEOS wrapper validates results from GEOS.

This feels like a misalignment:

- Validation isn't free, so validating GEOS results seems like the wrong default.

- The GEOS wrapper is doing more than strictly wrapping the GEOS library (it is also validating the result). This blurs the lines between what simplefeatures is doing and what GEOS is doing.

This change removes that validation. Users can of course still validate manually by calling the `Validate` method on the result.

## Check List

Have you:

- Added unit tests? No, because it's removal of an existing feature.

- Updated release notes? (if appropriate?) Not yet. There will be a large amount of changes for this release, so I'll add release notes at the end.

## Related Issue

## Benchmark Results

  `run_benchmarks.sh` script.

Not much to see here. All benchmark results are the same, except for benchmarks for GEOS operations which are now faster (due to the change in validation behaviour).

[bench_results.txt](https://github.com/peterstace/simplefeatures/files/12614454/bench_results.txt)

### DIFF
--- a/geos/benchmark_test.go
+++ b/geos/benchmark_test.go
@@ -33,23 +33,6 @@ func regularPolygon(center geom.XY, radius float64, sides int) geom.Polygon {
 	return poly
 }
 
-func BenchmarkIntersectionWithoutValidation(b *testing.B) {
-	for _, sz := range []int{10, 100, 1000, 10000} {
-		b.Run(fmt.Sprintf("n=%d", sz), func(b *testing.B) {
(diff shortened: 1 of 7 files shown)


albertteoh (approved):

peterstace:

PR #527, opened 2023-09-15
Remove constructor options from TransformXY methods

peterstace:
## Description

The `TransformXY` methods transforms the coordinates of geometries in arbitrary ways via a user-supplied mapping. It should be obvious that this potentially could cause invalid results. For example, the mapping could trivially map every coordinate pair of a polygon to the same point. In practice, transformations are either affine (translation, rotation, scaling), or, at the very least, topologically preserving (e.g. WGS84 to a Mercator or Conic projection).

This change removes the choice to validate or not in `TransformXY` methods, always omitting a validation step. Users can still validate manually via the `Validate` method if they wish to.

## Check List

Have you:

- Added unit tests? No, however existing tests have been updated.

- Updated release notes? (if appropriate?) No, release notes will be updated when all ctor option related changes have been made.

## Related Issue

### DIFF
--- a/geom/type_envelope.go
+++ b/geom/type_envelope.go
@@ -281,13 +281,12 @@ func (e Envelope) Distance(o Envelope) (float64, bool) {
 }
 
 // TransformXY transforms this Envelope into another Envelope according to fn.
-// TODO: remove the error here as well.
-func (e Envelope) TransformXY(fn func(XY) XY) (Envelope, error) {
+func (e Envelope) TransformXY(fn func(XY) XY) Envelope {
 	min, max, ok := e.MinMaxXYs()
 	if !ok {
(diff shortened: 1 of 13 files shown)


geom/type_envelope.go
```diff
@@ -281,6 +281,7 @@ func (e Envelope) Distance(o Envelope) (float64, bool) {
 }
 
 // TransformXY transforms this Envelope into another Envelope according to fn.
+// TODO: remove the error here as well.
```
peterstace, line 284: Woops, missed this one before raising a PR.

- [ ] Remove error from `Envelope`'s `TransformXY` signature.

albertteoh (approved):

peterstace:

PR #530, opened 2023-09-20
Remove error and ctor options from direct ctors

peterstace:
## Description

Modify direct constructors (`NewPoint` etc.) to no longer perform validation at all. The direct constructors are for more advanced use cases where geometries are being constructed from other geometries or raw sequences of floats. Things like custom geometric algorithms come to mind. Because these are for advanced use cases, making users validate manually is reasonable.

Modifying direct constructors removes error handling mismatch, with users no longer having to handle errors that can never occur (when validation is disabled). It also eliminates the situation where constructor options would sometimes be ignored, and simplifies the rules around what validation occurs during direct construction (i.e. now none is performed).

See https://github.com/peterstace/simplefeatures/discussions/525 for more context.

## Check List

Have you:

- Added unit tests? Yes.

- Updated release notes? (if appropriate?) No. This will be done when all constructor option changes are complete.

## Related Issue

## Benchmark Results

  `run_benchmarks.sh` script.

<details>

```
PASTE BENCHMARKS HERE
```

</details>

### DIFF
--- a/geom/accessor_test.go
+++ b/geom/accessor_test.go
@@ -25,13 +25,11 @@ func TestLineStringAccessor(t *testing.T) {
 	pt56 := xyCoords(5, 6)
 
 	t.Run("start", func(t *testing.T) {
-		want, err := NewPoint(pt12)
-		expectNoErr(t, err)
+		want := NewPoint(pt12)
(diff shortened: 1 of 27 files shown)


albertteoh (approved):

peterstace:

PR #554, opened 2023-09-29
Fix bug in Envelope's TransformXY method

peterstace:
## Description

The bug occurs when the transform function doesn't preserve the property that the envelopes min and max points remain the min and max points after the transformation. There is no expectation that this transform will preserve this property (e.g. WGS84 to WebMercator doesn't preserve the ordering due to the inverted Y-axis of WebMercator pixel space).

The fix is to reorder the min/max before constructing the new envelope.

## Check List

Have you:

- Added unit tests? Yes.

## Related Issue

- N/A

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,7 +19,7 @@ jobs:
     - name: cmpgeos
       run: make cmpgeos
     - name: Convert coverage to lcov
-      uses: jandelgado/gcov2lcov-action@v1.0.6
+      uses: jandelgado/gcov2lcov-action@v1.0.9
       with:
         infile: coverage.out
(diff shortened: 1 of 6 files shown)


.github/workflows/ci.yml
```diff
@@ -19,7 +19,7 @@ jobs:
     - name: cmpgeos
       run: make cmpgeos
     - name: Convert coverage to lcov
-      uses: jandelgado/gcov2lcov-action@v1.0.2
+      uses: jandelgado/gcov2lcov-action@v1.0.9
```
peterstace, line 22: Needed to get the latest gcov2lcov since the older one isn't aware of the retract directive.

PR #555, opened 2023-09-30
Add `Envelope` method to `Sequence`

peterstace:
## Description

The motivation behind this change is to allow Envelopes for linear and areal types to be calculated more efficiently. Rather than expanding an envelope XY by XY, an Envelope can be calculated directly from a sequence. This change will be made in a separate PR.

## Check List

Have you:

- Added unit tests? Yes.

## Related Issue

- N/A

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 YYYY-MM-DD
 
-- Adds an method with signature `Envelope() Envelope` to type `Sequence`.
+- Adds a method with signature `Envelope() Envelope` to type `Sequence`.
 
 ## v0.45.1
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+YYYY-MM-DD
+
+- Adds an method with signature `Envelope() Envelope` to type `Sequence`.
```
albertteoh, line 7:
```suggestion
- Adds a method with signature `Envelope() Envelope` to type `Sequence`.
```

albertteoh (approved):
👍🏼

peterstace:

PR #556, opened 2023-09-30
Simplify internal representation of Envelope

peterstace:
## Description

This change removes the "XOR NaN" hack. The hack was initially introduced as a performance optimisation, but is fairly complicated.

NOTE: this PR introduces a performance regression of up to 5% for some operations. The performance regression will be fixed in a subsequent PR my modifying how the `Envelope` method for `LineString`s is implemented.

## Check List

Have you:

- Added unit tests? Relies on existing.

## Related Issue

- N/A

## Benchmark Results

[results.txt](https://github.com/peterstace/simplefeatures/files/12785217/results.txt)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ YYYY-MM-DD
 
 - Adds an method with signature `Envelope() Envelope` to type `Sequence`.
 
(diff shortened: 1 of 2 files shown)


albertteoh (approved):

peterstace:
Thanks for reviewing!

PR #559, opened 2023-10-05
Better NaN handling for fastMin and fastMax

peterstace:
## Description

The `fastMin` and `fastMax` functions previously didn't taken NaN into account, so returned inaccurate results when NaNs were passed in, biasing NaN to be returned only if it were the first argument.

This change makes the behaviour of the functions symmetrical, returning NaN if _either_ argument (or both) are NaN.

Benchmarks have been added to show that indeed the fast variants are significantly faster than their standard variants.

```
$ go test -run=XXX -bench='Fast|Math' ./geom
goos: linux
goarch: arm64
pkg: github.com/peterstace/simplefeatures/geom
BenchmarkFastMin-2      712878825                1.574 ns/op
BenchmarkFastMax-2      1000000000               1.101 ns/op
BenchmarkMathMin-2      149962322                7.861 ns/op
BenchmarkMathMax-2      152543700                7.866 ns/op
PASS
```

The motivation behind this change is to remove the error from Envelope constructors, and perform validation separately as its own method (similar to the recent changes from geometries). Since the envelope only stores the min and max points, I'd like any NaNs to "infect" the envelope so that they're caught by validation rather than silently being dropped.

## Check List

Have you:

- Added unit tests? Yes.

## Related Issue

- N/A

### DIFF
--- a/geom/util.go
+++ b/geom/util.go
@@ -2,6 +2,7 @@ package geom
 
 import (
 	"fmt"
+	"math"
 	"sort"
 )
 
@@ -69,7 +70,7 @@ func sequenceToXYs(seq Sequence) []XY {
 
 // fastMin is a faster but not functionally identical version of math.Min.
 func fastMin(a, b float64) float64 {
-	if a < b {
+	if math.IsNaN(a) || a < b {
 		return a
 	}
 	return b
@@ -77,7 +78,7 @@ func fastMin(a, b float64) float64 {
 
 // fastMax is a faster but not functionally identical version of math.Max.
 func fastMax(a, b float64) float64 {
-	if a > b {
+	if math.IsNaN(a) || a > b {
 		return a
 	}
 	return b
--- a/geom/util_internal_test.go
+++ b/geom/util_internal_test.go
@@ -1,0 +1,71 @@
+package geom
+
+import (
+	"math"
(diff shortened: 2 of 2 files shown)


geom/util.go
```diff
@@ -69,15 +70,15 @@ func sequenceToXYs(seq Sequence) []XY {
 
 // fastMin is a faster but not functionally identical version of math.Min.
 func fastMin(a, b float64) float64 {
-	if a < b {
+	if math.IsNaN(a) || a < b {
```
albertteoh, line 73: Nice, looks to also be consistent behaviour with the built-in `min` (and `max`) funcs offered in Go 1.21.

Do we need to handle `+-Inf` as well?

albertteoh (approved):

geom/util.go
peterstace: Yes, once I upgrade to go 1.21, we _might_ not need these anymore. Probably not going to upgrade though until Go 1.23 comes out, in order to not force users to upgrade early.

It doesn't need to handle infinity _explicitly_, since it's handled implicitly by the `a < b` check. All of the cases involving infinity "just work" 😁.

peterstace:
Thanks for reviewing @albertteoh 🙇

Issue #522, opened 2023-09-05
Update golangci-lint to use latest version


peterstace:
The latest version of golangci-lint has breaking changes for depguard.

Because I'm running the latest version of golangci-lint locally (and it's hard to downgrade), this makes things a bit annoying.

AC:

- [ ] Update CI to use the latest golangci-lint. This is 1.54.2 at the time of writing, but may be higher by the time this is worked on.
- [ ] Review/revisit the golangci-lint configuration that we have, to make sure there isn't anything extraneous there.

PR #570, opened 2023-11-12
Upgrade golangci-lint to v1.55.2

peterstace:
## Description

Some of the changes involve the `revive` linter being a bit stricter, resulting in the removal of "dot" imports. This is probably a good thing, since the tests are now more similar to how users will see things. There is also a new linter used called `perfsprint` which checks for bad uses of `fmt.Sprintf`.

## Check List

Have you:

- Added unit tests? N/A

## Related Issue

- https://github.com/peterstace/simplefeatures/issues/522

### DIFF
--- a/.ci/golint.Dockerfile
+++ b/.ci/golint.Dockerfile
@@ -1,4 +1,4 @@
-FROM golangci/golangci-lint:v1.54.2
+FROM golangci/golangci-lint:v1.55.2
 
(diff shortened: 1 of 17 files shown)


albertteoh (approved):
Nice 👍🏼

peterstace:
Thanks  for reviewing @albertteoh 😁

Issue #569, opened 2023-10-24
No rings error on intersection error


angaither:
I think we have a faulty dcel ring extraction. 

Given two valid geometries, an intersection should return. 
Here is a test that should pass in both cases against the latest tag [v0.45.1](https://github.com/peterstace/simplefeatures/releases/tag/v0.45.1) however the first example fails due to "no rings to extract" 
When the geometry is simplified to 6 floating decimal points (second test case) an intersection is returned. 

I believe it is related to the issue mentioned in [this commit](https://github.com/peterstace/simplefeatures/commit/5e2af2fd5e84e70c505f43851a066efca5b0745c)

I added the following test to `geom/alg_set_op_test.go`
```go
func TestIntersection(t *testing.T) {
	for i, tc := range []struct {
		inputA, inputB, want string
	}{
		{
			// failing case with no rings to extract (should not fail).
			"POLYGON ((-57.84764391579377 -14.00436771429812, -57.98105430423379 -13.978568346975345, -57.97219 -13.895754, -57.815573 -13.870471, -57.78975494169227 -13.97408746357712, -57.79567678742665 -14.003207561112367, -57.84764391579377 -14.00436771429812))",
			"POLYGON ((-57.97219 -13.895754, -57.815573 -13.870471, -57.782572 -14.002915, -57.984142 -14.007415, -57.97219 -13.895754))",
			"POLYGON((-57.97219 -13.895754,-57.98105430423379 -13.978568346975345,-57.84764391579377 -14.00436771429812,-57.82788805023705 -14.003926669524565,-57.79567678742665 -14.003207561112367,-57.78975494169227 -13.97408746357712,-57.815573 -13.870471,-57.97219 -13.895754))",
		},
		{
			// simplified, same coords to 6 decimal points.
			"POLYGON ((-57.847643 -14.004367, -57.981054 -13.978568, -57.97219 -13.895754, -57.815573 -13.870471, -57.789754 -13.974087, -57.795676 -14.003207, -57.847643 -14.004367))",
			"POLYGON ((-57.97219 -13.895754, -57.815573 -13.870471, -57.782572 -14.002915, -57.984142 -14.007415, -57.97219 -13.895754))",
			"POLYGON((-57.97219 -13.895754,-57.981054 -13.978568,-57.847643 -14.004367,-57.795676 -14.003207,-57.78975447509755 -13.974089336177082,-57.815573 -13.870471,-57.97219 -13.895754))",
		},
	} {
[...]

PR #571, opened 2023-11-21
Separate point/line and line/line renoding in DCEL

peterstace:
## Description

Previously, the point/line and line/line renoding steps were intertwined during DCEL geometry renoding. This caused bad DCEL structures to be produced in some real-world inputs.

Logically, point/line renoding should occur before line/line renoding.  This is because point/line renoding can snap together overlapping line segments. When line segments completely overlap, they are skipped by the line/line renoding step. If they are not snapped together, then numeric precision issues in the line/line renoding stage can cause DCEL points that are very close together. By doing point/line renoding first, the entire problem is avoided.

## Check List

Have you:

- Added unit tests? Yes.

- Updated release notes? (if appropriate?) Yes

## Related Issue

- https://github.com/peterstace/simplefeatures/issues/569

## Benchmark Results

There are some 10% to 40% perf degradations for DCEL operations. Unfortunate, but in the name of a bug fix I don't think there's much choice.

[bench.txt](https://github.com/peterstace/simplefeatures/files/13437657/bench.txt)


albertteoh (approved):
The changes are beyond my realm of understanding tbh, but the addition of a reproducer test and implementation to pass it are good enough for me! 😄

peterstace:
Thanks for looking @albertteoh 😁

Issue #574, opened 2023-11-22
Use symmetric line/line intersection


peterstace:
While debugging https://github.com/peterstace/simplefeatures/issues/569, I stumbled across an unrelated bug.

The line vs. line intersection method is (unexpectedly not symmetric, i.e. doesn't produce the same output when its inputs are reversed. Because the line vs. line intersection calculation is performed once in each direction for each unique pair of lines that could intersect, re-nodding anomalies can occur if the two results are _sufficiently_ different.

There are two possible solutions:

1. During the line vs. line intersection calculation, first reorder the inputs in a deterministic way.
2. Use a calculation that is completely symmetric (which may be difficult since there are a lot of float64 operations involved, which aren't inherently symmetric).

This issue is _potentially_ the root cause of others, e.g.:

- https://github.com/peterstace/simplefeatures/issues/573
- https://github.com/peterstace/simplefeatures/issues/496

Issue #573, opened 2023-11-22
geom.Union failed at some cases


dongzhou77:
Hey, 

geom.Union failed at some cases.

Input geometries:
geometry a:
[...]

PR #575, opened 2023-11-23
Hack: swap line vs. line intersection order

peterstace:
## Description

This fixes a bug where anomalously close DCEL nodes cause DCEL extraction to fail. The bug would occur because the line vs. line intersection operation was not symmetric, and was called from "both directions" when re-noding input geometries.

This solution is a bit of a hack. Instead, a longer term solution is to truly make the operation symmetric (e.g. by implementing a more numerically stable algorithm).

## Check List

Have you:

- Added unit tests? Yes.

## Related Issue

- https://github.com/peterstace/simplefeatures/issues/574
- https://github.com/peterstace/simplefeatures/issues/573
- https://github.com/peterstace/simplefeatures/issues/496

## Benchmark Results

TODO


albertteoh (approved):

peterstace:
Thanks for reviewing Albert! I'm going to have a crack at a better solution for this a bit later (i.e. to go into the next release).

Issue #558, opened 2023-10-04
Remove unneeded benchmarks


peterstace:
The benchmark suite currently takes a very long time to execute. Many of the benchmarks are just slight variations of each other. This issue is to evaluate which benchmarks are useful, and get rid of the rest.

PR #577, opened 2023-11-26
Remove/reduce unnecessary benchmarks

peterstace:
## Description

- Removes `BenchmarkWKBParse`, which is actually ~70 tests. There is already a `BenchmarkUnmarshalWKB` in `geom/perf_test.go`, which not being as comprehensive, will alert developers to unexpected unmarshal performance regressions.

- Alters `BenchmarkSetOperation` to have 12 sub benchmarks instead of 48.  This reduces the granularity of the sizes of inputs tested, but still gives reasonable coverage.

## Check List

Have you:

- Added unit tests? N/A

## Related Issue

https://github.com/peterstace/simplefeatures/issues/558


albertteoh (approved):
👍🏼

peterstace:

Issue #531, opened 2023-09-20
Unify GEOS wrappers


peterstace:
There are currently two separate GEOS wrappers:

1. in the `github.com/peterstace/simplefeatures/geos` package.
2. in the `github.com/peterstace/simplefeatures/internal/cmpgeos` package.

We should be able to refactor things so there is only 1 set of wrappers.

PR #578, opened 2023-11-28
Move low level details `geos` package into `internal/rawgeos`

peterstace:
## Description

This is to pave the way to allow unification of the GEOS wrapper between the `geos` `cmprefimpl/cmpgeos` packages. The idea is that the `internal/rawgeos` package will (soon) be a dependency of both packages, with the extra wrapped functions that `cmpgeos` requires (soon) moved from `cmpgeos` into `internal/rawageos`.

## Check List

Have you:

- Added unit tests? N/A

- Updated release notes? (if appropriate?) Not yet -- I'll update these once the geos/cmpgeos packages are fully unified.

## Related Issue

- https://github.com/peterstace/simplefeatures/issues/531


peterstace:

PR #579, opened 2023-11-29
Remove `cmpgeos` wrappers that already exist in the `rawgeos` package

peterstace:
## Description

Since Intersects, Union etc. are already wrapped in the `rawgeos` package, they can be used instead, and the corresponding wrappers in the `cmpgeos` package deleted.

The next step will be to _move_ the remaining wrappers from the `cmpgeos` package to the `rawgeos` package.

## Check List

Have you:

- Added unit tests? N/A

- Updated release notes? (if appropriate?) No, will write release notes when the cmpgeos wrapper cleanup is complete.

## Related Issue

- https://github.com/peterstace/simplefeatures/issues/531


albertteoh (approved):
Nice

peterstace:
Thanks for reviewing, I really appreciate it 🥳

Issue #561, opened 2023-10-08
Use alpine images for GEOS testing


peterstace:
Images are currently debian images that have GEOS installed manually inside them. We may be able to use alpine images instead so that we don't have to cross-compile.

PR #581, opened 2023-12-17
Use alpine images for GEOS tests

peterstace:
## Description

The previous approach compiled the GEOS library from source. This was quite annoying when new versions of GEOS came out, because new images had to be created (they take several hours to build on a powerful machine).

The new approach is simpler and is based on pre-built alpine images.

## Check List

Have you:

- Added unit tests? N/A

## Related Issue

- https://github.com/peterstace/simplefeatures/issues/561


geos/entrypoints_test.go
peterstace, line 1: The removed test cases in this file no longer pass with GEOS 3.11 (they were a little bit narrow focused).

albertteoh (approved):
Why was the approach of compiling the geos library from source taken initially?

peterstace:
> Why was the approach of compiling the geos library from source taken initially?

I thought it was important to be able to test against old versions of GEOS, but incorrectly thought that was impossible/hard to do with a pre-compiled version of the library. I mistakenly believed that alpine was a rolling release, where only the latest version of everything is available. In reality, you can pin to a specific alpine release. Other distros (e.g. debian) have major releases at such a slow rate that some of the minor versions of GEOS are skipped between OS releases.

(btw, thanks for reviewing!)

PR #585, opened 2024-01-20
Use `XY.Less` comparison func

peterstace:
## Description
    
`XY` has `Less` method to compare two `XY` values. This logic was duplicated in two other places. This change removes the duplication, instead relying on the `Less` method.

## Check List

Have you:

- Added unit tests? N/A

- Updated release notes? (if appropriate?) N/A (change too small to both with release notes)

## Related Issue

- N/A


albertteoh (approved):
👍🏼

peterstace:
Thanks for reviewing @albertteoh! 😁

PR #588, opened 2024-01-30
Implement Densify methods

peterstace:
## Description

These methods return copies of the input geometry that have additional control points inserted such that adjacent control points are at most a fixed distance apart.

This is useful when transforming geometries from one projection to another, since straight lines in one projection are not necessarily straight lines in other projections. The additional control points can help to retain the true shape of curves when they are transformed.

## Check List

Have you:

- Added unit tests? Yes.

## Related Issue

- N/A

### DIFF
--- a/geom/alg_densify_test.go
+++ b/geom/alg_densify_test.go
@@ -73,3 +73,22 @@ func TestDensify(t *testing.T) {
 		})
 	}
 }
+
+func TestDensifyInvalidMaxDist(t *testing.T) {
+	for i, tc := range []struct {
+		input   string
+		maxDist float64
+	}{
+		{"LINESTRING(0 0,1 0)", -1},
+		{"LINESTRING(0 0,1 0)", 0},
+		{"POINT(0 0)", -1},
+		{"POINT(0 0)", 0},
+		{"MULTIPOINT((0 0))", -1},
+		{"MULTIPOINT((0 0))", 0},
+	} {
+		t.Run(strconv.Itoa(i), func(t *testing.T) {
+			input := geomFromWKT(t, tc.input)
+			expectPanics(t, func() { input.Densify(tc.maxDist) })
+		})
+	}
+}
--- a/geom/alg_densify_test.go
+++ b/geom/alg_densify_test.go
@@ -1,0 +1,75 @@
+package geom_test
+
+import (
+	"strconv"
+	"testing"
+
+	"github.com/peterstace/simplefeatures/geom"
+)
+
+func TestDensifyEmpty(t *testing.T) {
+	for _, empty := range []geom.Geometry{
+		geom.Point{}.AsGeometry(),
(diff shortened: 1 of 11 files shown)


geom/alg_densify_test.go
```diff
@@ -0,0 +1,75 @@
...
+				})
+			}
+		})
+	}
+}
+
+func TestDensify(t *testing.T) {
+	for i, tc := range []struct {
+		input   string
+		minDist float64
```
albertteoh, line 40: Should this be `maxDist`?

geom/alg_densify_test.go
```diff
@@ -0,0 +1,75 @@
...
+		want    string
+	}{
+		// LineString with a single segment (tests threshold logic):
+		{"LINESTRING(0 0,1 0)", 2.0, "LINESTRING(0 0,1 0)"},
+		{"LINESTRING(0 0,1 0)", 1.0, "LINESTRING(0 0,1 0)"},
+		{"LINESTRING(0 0,1 0)", 0.9, "LINESTRING(0 0,0.5 0,1 0)"},
+		{"LINESTRING(0 0,1 0)", 0.5, "LINESTRING(0 0,0.5 0,1 0)"},
+		{"LINESTRING(0 0,1 0)", 0.4, "LINESTRING(0 0,0.3333333333333333 0,0.6666666666666666 0,1 0)"},
+		{"LINESTRING(0 0,1 0)", 0.3, "LINESTRING(0 0,0.25 0,0.5 0,0.75 0,1 0)"},
+		{"LINESTRING(0 0,1 0)", 0.25, "LINESTRING(0 0,0.25 0,0.5 0,0.75 0,1 0)"},
```
albertteoh, line 50: Is having a `maxDist = 0` a realistic scenario that we should test/guard for?

albertteoh (approved):
👍🏼

geom/alg_densify_test.go
peterstace: Yes! Fixed in 557a096cf6b2ea30bd306128146fd81ac2f03cad

geom/alg_densify_test.go
peterstace: Ah! A user supplying weird values of `maxDist` is something I hadn't considered.

If a user supplies zero, then they will get an infinite loop. If a user provides a _very small_ max dist, the resultant geometry will just be very dense (it's easy to create a geometry with millions of control points). If a user supplies a negative max dist, then no additional control points are added (which feels weird).

We don't want to disallow very small but non-zero values of `maxDist artificially` (it would be too hard to get a reasonable cutoff in all use cases). However, we want to disallow zero and negative values.

557a096cf6b2ea30bd306128146fd81ac2f03cad detects the negative/zero cases and panics. That's reasonable behaviour since it would be a programmer error to supply a negative/zero value.

PR #598, opened 2024-03-18
Add ConcaveHull from geos lib

valri11:
## Description

Add ConcaveHull from geos lib

### DIFF
--- a/geos/entrypoints.go
+++ b/geos/entrypoints.go
@@ -286,3 +286,15 @@ func CoverageUnion(g geom.Geometry) (geom.Geometry, error) {
 func UnaryUnion(g geom.Geometry) (geom.Geometry, error) {
 	return rawgeos.UnaryUnion(g)
 }
+
+// ConcaveHull returns concave hull of input geometry.
+// pctconvex - ratio 0 to 1 (0 - max concaveness, 1 - convex hull)
+// allowHoles - true to allow holes inside of polygons
+func ConcaveHull(g geom.Geometry, pctconvex float64, allowHoles bool) (geom.Geometry, error) {
+	return rawgeos.ConcaveHull(g, pctconvex, allowHoles)
+}
+
+// ConvexHull returns convex hull of input geometry
+func ConvexHull(g geom.Geometry) (geom.Geometry, error) {
+	return rawgeos.ConvexHull(g)
+}
--- a/.ci/alpine_geos.pc
+++ b/.ci/alpine_geos.pc
@@ -1,0 +1,10 @@
+prefix=/usr
+exec_prefix=${prefix}
+libdir=${exec_prefix}/lib
+includedir=${prefix}/include
+
+Name: GEOS
+Description: Geometry Engine - Open Source
+Version: <version>
+Libs: -L${libdir} -lgeos_c
+Cflags: -I${includedir}
--- a/.ci/geos.Dockerfile
+++ b/.ci/geos.Dockerfile
@@ -2,7 +2,12 @@ ARG ALPINE_VERSION
 FROM alpine:${ALPINE_VERSION}
 
 ARG GEOS_VERSION
-RUN apk add gcc musl-dev geos-dev=${GEOS_VERSION}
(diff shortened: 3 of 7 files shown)


peterstace (approved):
Thanks @valri11, this looks good, I'll merge soon.

peterstace:
Hi @valri11, I attempted to run the CI jobs against your branch, but some of the tests are failing. I'm still not sure why CI is not running for your branch, but you should be able to reproduce the issues locally. The failing commands are:

```
make lint
make goes
```

The `make lint` errors look pretty trivial to fix.

The `make geos` errors might be a bit harder. They're failing because `GEOSConcaveHull_r` was only added in GEOS 3.11, which prevents the `github.com/peterstace/simplefeatures/geos` package compiling against earlier versions. You may be able to solve this by using the same approach as `GEOSMakeValid_r` and `GEOSCoverageUnion_r`, which were only introduced in GEOS 3.8.

peterstace:
@valri11 , I've altered the github actions config in https://github.com/peterstace/simplefeatures/pull/601, which should hopefully resolve the issue where actions are not running against your fork. For actions to run, I _think_ you'll need to:

- Push another commit to trigger it.
- Then I'll need to (once-off) approve your fork (as per https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks).

peterstace (changes requested):
Changing status to "Request changes" due to failing CI.

geos/entrypoints.go
```diff
@@ -286,3 +286,15 @@ func CoverageUnion(g geom.Geometry) (geom.Geometry, error) {
...
 }
+
+// ConcaveHull returns concave hull of input geometry.
+// pctconvex - ratio 0 to 1 (0 - max concaveness, 1 - convex hull)
+// allowHoles - true to allow holes inside of polygons.
+func ConcaveHull(g geom.Geometry, pctconvex float64, allowHoles bool) (geom.Geometry, error) {
+	return rawgeos.ConcaveHull(g, pctconvex, allowHoles)
+}
+
+// ConvexHull returns convex hull of input geometry.
```
peterstace, line 297: Can you remove `ConvexHull`? Typically, I haven't been exposing functionality already implemented as part of the `github.com/peterstace/simplefeatures/geom` package (i.e. native Go implementation) unless there is a good reason to.

geos/entrypoints.go
```diff
@@ -286,3 +286,15 @@ func CoverageUnion(g geom.Geometry) (geom.Geometry, error) {
 func UnaryUnion(g geom.Geometry) (geom.Geometry, error) {
 	return rawgeos.UnaryUnion(g)
 }
+
+// ConcaveHull returns concave hull of input geometry.
+// pctconvex - ratio 0 to 1 (0 - max concaveness, 1 - convex hull)
+// allowHoles - true to allow holes inside of polygons.
+func ConcaveHull(g geom.Geometry, pctconvex float64, allowHoles bool) (geom.Geometry, error) {
```
peterstace, line 293: Nit: could you rename `pctconvex` to either `ratio` or maybe `convexityRatio`? The `pct` prefix isn't really right because it's a ratio (0 to 1) rather than a percentage (0 to 100). As a reference, the GEOS interface uses `ratio` as the name.

peterstace (changes requested):
Just two more request, then I think this will be good to merge.

geos/entrypoints.go
valri11: added here as a pair to ConcaveHull. Could be useful for comparing performance of pure go implementation. Anyway, will remove it - ConcaveHull with ratio=1 is ConvexHull

geos/entrypoints.go
valri11: I found ratio is confusing here, param named similar to https://postgis.net/docs/ST_ConcaveHull.html

valri11: addressed review comments

peterstace (approved):
Amazing, thanks @valri11 !

PR #601, opened 2024-04-03
Trigger CI when PRs are created/updated

peterstace:
## Description

Right now, CI is not run when PRs are raised against forks. By adding `pull_request` as a triggering action, I *think* this will cause them to run for forks as well (this is a little bit experimental).

References:

- https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#push

- https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request

## Check List

Have you:

- Added unit tests? N/A

## Related Issue

N/A

## Benchmark Results

N/A

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,6 +1,7 @@
 name: build
 on:
   push:
+  pull_request:
 jobs:
   build:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


peterstace:
Hmm, this seems to be running the action _twice_, which is not what I intended...

Screenshot 2024-04-04 at 5 29 40 am

PR #594, opened 2024-03-08
Use pkg-config for configuring LDFLAGS

peterstace:
## Description

This is to allow the `github.com/peterstace/geos` package to compile on new M2 Macs.

## Check List

Have you:

- Added unit tests? N/A

- Add cmprefimpl tests? (if appropriate?) N/A

- Updated release notes? (if appropriate?) N/A

## Related Issue

- N/A

PR #602, opened 2024-04-11
Allow the use of `pkg-config` to be opt-out

peterstace:
## Description

Some systems don't have `pkg-config`, or have `pkg-config` that's set up
incorrectly. These systems couldn't easily use the `v0.47.1` tag,
because that tag introduces a hard dependency on `pkg-config` being used
to configure GEOS.

This changeset allows `pkg-config` to be opted out of, by using the
`sfnopkgconfig` build tag. The behaviour when using that build tag is
effectively the same as pre-v0.47.1.

## Check List

Have you:

- Added unit tests? Yes.

## Related Issue

- Relates to https://github.com/peterstace/simplefeatures/pull/594


internal/rawgeos/entrypoints.go
peterstace, line 1: The `#cgo` directives only need to be applied _once_, and are now supplied by either `configure_hardcoded.go` or `configure_pkg_config.go` (depending on if the `sfnopkgconfig` build tag is set or not).

PR #611, opened 2024-05-07
Upgrade to golangci-lint v1.58.0

peterstace:
## Description

This jumps up two major versions and a few minor versions, so there are a few changes (I did the upgrade one version at a time).

## Check List

Have you:

- Added unit tests? N/A

## Related Issue

- N/A

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -42,6 +42,7 @@ linters:
     - bodyclose
     - containedctx
     - contextcheck
+    - copyloopvar
     - decorder
     - depguard
     - dogsled
@@ -50,7 +51,6 @@ linters:
     - errcheck
     - errchkjson
     - errorlint
-    - execinquery
     - exportloopref
     - ginkgolinter
     - gocheckcompilerdirectives
@@ -71,6 +71,7 @@ linters:
     - importas
     - ineffassign
     - interfacebloat
+    - intrange
     - ireturn
     - loggercheck
     - makezero
@@ -112,6 +113,7 @@ linters:
 
     # Deprecated by golangci-lint:
     - deadcode
+    - execinquery
     - exhaustivestruct
     - golint
     - ifshort
@@ -140,7 +142,7 @@ linters:
     - goconst
     - gocyclo
     - godox
-    - goerr113
+    - err113
     - gomnd
     - inamedparam
     - lll
--- a/geom/dcel_debug.go
+++ b/geom/dcel_debug.go
@@ -6,7 +6,7 @@ import (
 	"sort"
 )
 
-//nolint:deadcode,unused
+//nolint:unused
 func dumpDCEL(d *doublyConnectedEdgeList) {
 	newNamedDCEL(d).show()
 }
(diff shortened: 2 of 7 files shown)


.golangci.yaml
```diff
@@ -42,6 +42,7 @@ linters:
     - bodyclose
     - containedctx
     - contextcheck
+    - copyloopvar
```
peterstace, line 45: `copyloopvar` and `intrange` are enabled, but just give a warning that they don't do anything unless using at least Go 1.22 (we're still using Go 1.17). We definitely don't want to upgrade beyond the features we need, since that would needlessly require users to upgrade.

.golangci.yaml
```diff
@@ -112,6 +113,7 @@ linters:
 
     # Deprecated by golangci-lint:
     - deadcode
+    - execinquery
```
peterstace, line 116: `execinquery` has become deprecated.

.golangci.yaml
```diff
@@ -140,7 +142,7 @@ linters:
     - goconst
     - gocyclo
     - godox
-    - goerr113
+    - err113
```
peterstace, line 145: `goerr113` was renamed by golangci-lint.

geom/dcel_debug.go
```diff
@@ -6,7 +6,7 @@ import (
 	"sort"
 )
 
-//nolint:deadcode,unused
+//nolint:unused
```
peterstace, line 9: `deadcode` was already disabled, so the `nolint` directive didn't need to include it.

geom/geojson_feature_collection.go
```diff
@@ -153,7 +153,7 @@ func (c *GeoJSONFeatureCollection) UnmarshalJSON(p []byte) error {
 		Type     string           `json:"type"`
 		Features []GeoJSONFeature `json:"features"`
 	}
-	if err := json.Unmarshal(p, &topLevel); err != nil {
+	if err := json.Unmarshal(p, &topLevel); err != nil { //nolint:musttag
```
peterstace, line 156: I believe there's a false positive bug in `musttag` that's triggered by this code.

I will try to create a minimum reproduction of the problem and, if successful, will raise a bug report.

geom/xy.go
```diff
@@ -71,7 +71,7 @@ func (w XY) Cross(o XY) float64 {
 	// Avoid fused multiply-add by explicitly converting intermediate products
 	// to float64. This ensures that the cross product is *exactly* zero for
 	// all linearly dependent inputs.
-	return float64(w.X*o.Y) - float64(w.Y*o.X) //nolint:unconvert
+	return float64(w.X*o.Y) - float64(w.Y*o.X)
```
peterstace, line 74: `unconvert` doesn't give a false positive here anymore 😁

geom/geojson_feature_collection.go
peterstace: Reported here: https://github.com/go-simpler/musttag/issues/93

albertteoh (approved):
👍🏼

peterstace:
Thanks for reviewing @albertteoh ! 😁

Issue #397, opened 2021-07-31
Investigate mixed coordinate types with GeometryCollection


peterstace:
E.g. the following WKT parses as valid: `GEOMETRYCOLLECTION M(POINT Z(1 2 0))`. But it's not internally consistent with respect to coordinate type (Z vs M).

We should give an error in these sorts of cases.

PR #613, opened 2024-05-13
Bugfix: check geometry collection coordinate types

peterstace:
## Description

Fixes a bug where geometry collection coordinate types were allowed to mismatch. E.g. `GEOMETRYCOLLECTION M(POINT Z(1 2 0))` is not allowed because the collection is marked as `M` but the point is marked as `Z`.

An extension beyond the OGC standard is implemented, allowing untagged geometry collections to contain any coordinate types, so long as they are all the same. So `GEOMETRYCOLLECTION(POINT M(1 2 0), POINT M(2 3 0))` is allowed, but `GEOMETRYCOLLECTION(POINT Z(1 2 0), POINT M(2 3 4))` is not.

## Check List

Have you:

- Added unit tests? Yes.

## Related Issue

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@
 
 - Upgrades `golangci-lint` to `v1.58.0`.
 
+- Fixes a bug where geometry collections with mixed coordinate types were
+  erroneously allowed during WKT parsing.
+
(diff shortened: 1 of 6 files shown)


peterstace:
Thanks for reviewing!

PR #615, opened 2024-05-14
Add dimension check when parsing WKB GeometryCollections

peterstace:
## Description

Add dimension check when parsing WKB GeometryCollections
    
This ensures that coordinate types are consistent between the geometry collection and its children.

## Check List

Have you:

- Added unit tests? Yes.

## Related Issue

### DIFF
--- a/geom/wkb_test.go
+++ b/geom/wkb_test.go
@@ -3,6 +3,7 @@ package geom_test
 import (
 	"bytes"
 	"encoding/hex"
+	"fmt"
 	"strconv"
 	"strings"
 	"testing"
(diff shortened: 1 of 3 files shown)


geom/wkb_test.go
```diff
@@ -678,3 +679,46 @@ func TestWKBUnmarshalEndianess(t *testing.T) {
...
+					gc.hex +
+					u32One +
+					bigEndian +
+					point.hex +
+					strings.Repeat(f64Zero, point.ct.Dimension())
+				_, err := geom.UnmarshalWKB(hexStringToBytes(t, hex))
+				if gc.ct == point.ct {
+					expectNoErr(t, err)
+				} else {
+					expectErr(t, err)
```
albertteoh, line 719: Should we assert that the error type is `mismatchedGeometryCollectionDimsError`?

albertteoh (approved):

peterstace:
Thanks for reviewing!

geom/wkb_test.go
peterstace: Actually.... maybe there is a way to get the best of both worlds. I'll do some experimentation and raise a fix in a new PR if I can get it working.

peterstace: I tried something out in https://github.com/peterstace/simplefeatures/pull/616 (it seems to work pretty well, would love to get your opinion.

PR #616, opened 2024-05-15
Better error assertions in tests

peterstace:
## Description

This change introduces better error assertions, prompted by a comment from Albert Teoh here: https://github.com/peterstace/simplefeatures/pull/615#discussion_r1601290941

The change introduces two new test helper functions:

- `expectErrIs(tb testing.TB, err, want error)`

- `expectErrAs(tb testing.TB, err error, target interface{})`

These wrap `errors.Is` and `errors.As` functions from the standard library, respectively.

In order for this to work, the `mismatchedGeometryCollectionDimsError` must be exported, but _only during tests_ (because we don't want to expose it to users of the library). This is achieved by using type aliases in a `_test.go` file in the `geom` package. Because `_test.go` files are only included during tests, this is a safe way to expose the type only during tests.

## Check List

Have you:

- Added unit tests? Yes.

- Updated release notes? (if appropriate?) N/A (internal change, so probably doesn't need them).

## Related Issue

### DIFF
--- a/geom/errors.go
+++ b/geom/errors.go
@@ -59,11 +59,11 @@ func wrapWithGeoJSONSyntaxError(err error) error {
 }
 
 type mismatchedGeometryCollectionDimsError struct {
-	ct1, ct2 CoordinatesType
+	CT1, CT2 CoordinatesType
 }
 
 func (e mismatchedGeometryCollectionDimsError) Error() string {
-	return fmt.Sprintf("mixed dimensions in geometry collection: %s and %s", e.ct1, e.ct2)
+	return fmt.Sprintf("mixed dimensions in geometry collection: %s and %s", e.CT1, e.CT2)
 }
 
 type ruleViolation string
--- a/geom/export_test.go
+++ b/geom/export_test.go
@@ -1,0 +1,3 @@
+package geom
+
+type MismatchedGeometryCollectionDimsError = mismatchedGeometryCollectionDimsError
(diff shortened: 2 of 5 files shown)


geom/errors.go
```diff
@@ -59,11 +59,11 @@ func wrapWithGeoJSONSyntaxError(err error) error {
 }
 
 type mismatchedGeometryCollectionDimsError struct {
-	ct1, ct2 CoordinatesType
+	CT1, CT2 CoordinatesType
```
peterstace, line 62: These need to be exported, otherwise they can't be accessed from the `geom_test` package. Because `mismatchedGeometryCollectionDimsError` isn't exported (except during tests), having `CT1` and `CT2` exported doesn't make any difference for users of the library.

geom/export_test.go
```diff
@@ -0,0 +1,3 @@
+package geom
+
+type MismatchedGeometryCollectionDimsError = mismatchedGeometryCollectionDimsError
```
albertteoh, line 3: Wizard! TIL something new about Go 🙏🏼

albertteoh (approved):
Amazing! 👍🏼

geom/export_test.go
peterstace: I only learned about this yesterday 😁

peterstace:
Thanks for reviewing!

PR #617, opened 2024-05-19
Upgrade to golangci-lint v0.58.2

peterstace:
## Description

As per title.

## Check List

Have you:

- Added unit tests? N/A

## Related Issue

- N/A

### DIFF
--- a/geom/geojson_feature_collection.go
+++ b/geom/geojson_feature_collection.go
@@ -153,7 +153,7 @@ func (c *GeoJSONFeatureCollection) UnmarshalJSON(p []byte) error {
 		Type     string           `json:"type"`
 		Features []GeoJSONFeature `json:"features"`
 	}
-	if err := json.Unmarshal(p, &topLevel); err != nil { //nolint:musttag
+	if err := json.Unmarshal(p, &topLevel); err != nil {
 		return err
 	}
 	if topLevel.Type == "" {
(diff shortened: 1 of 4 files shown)


geom/geojson_feature_collection.go
```diff
@@ -153,7 +153,7 @@ func (c *GeoJSONFeatureCollection) UnmarshalJSON(p []byte) error {
 		Type     string           `json:"type"`
 		Features []GeoJSONFeature `json:"features"`
 	}
-	if err := json.Unmarshal(p, &topLevel); err != nil { //nolint:musttag
```
peterstace, line 156: `nolint:musttag` was to work around a bug, which was fixed in https://github.com/go-simpler/musttag/issues/93.

geom/type_null_geometry.go
```diff
@@ -24,7 +24,7 @@ func (ng *NullGeometry) Scan(value interface{}) error {
 // Value implements the database/sql/driver.Valuer interface.
 func (ng NullGeometry) Value() (driver.Value, error) {
 	if !ng.Valid {
-		return nil, nil
+		return nil, nil //nolint:nilnil
```
peterstace, line 27: `return nil, nil` is being detected as suspicious (usually this would be an error). It's not here, because the `database/sql/driver.Valuer` interface expects nil values (it interprets them as `NULL`).

albertteoh (approved):

peterstace:
Thanks for the review!

Issue #587, opened 2024-01-29
Coordinate type not retained for simplified empty geometries


peterstace:
E.g. https://go.dev/play/p/ESSvqvTFVlV

```
package main

import (
	"fmt"

	"github.com/peterstace/simplefeatures/geom"
)

func main() {
	p := geom.Polygon{}.ForceCoordinatesType(geom.DimXYM)
	p, err := p.Simplify(0.1)
	if err != nil {
		panic(err)
	}
	fmt.Println(p.CoordinatesType())
}
```
```
XY

Program exited.
```

PR #619, opened 2024-05-22
Retain coordinate type for empty simplified geometries

peterstace:
## Description

This fixes a bug where, in some scenarios, the `Simplify` operation would drop extended coordinate type information (from XYZ, XYM, or XYZM back to XY).
    
The bug happened because some geometry constructors inferred the geometry type from the input's lowest common denominator coordinate type. But if the input is empty, then the coordinate type is inferred as XY. The constructor design is a bit of a sharp edge and is not addressed here (instead, the bug that the sharp edge made more likely is fixed).

## Check List

Have you:

- Added unit tests? Yes.

## Related Issue

- https://github.com/peterstace/simplefeatures/issues/587


albertteoh (approved):

peterstace:
Thanks @albertteoh ! 😁

Issue #424, opened 2021-11-01
Wrap GEOSTopologyPreserveSimplify_r


peterstace:
We have regular simplify, but not the topological preserve variant. It may come in handy.

PR #621, opened 2024-05-30
Wrap `GEOSTopologyPreserveSimplify_r`

peterstace:
## Description

`TopologyPreserveSimplify` is similar to `Simplify` but preserves (where possible) topological invariants, e.g. ring collapse and intersection.

## Check List

Have you:

- Added unit tests? Yes.

## Related Issue

- Resolves https://github.com/peterstace/simplefeatures/issues/424

### DIFF
--- a/geos/entrypoints.go
+++ b/geos/entrypoints.go
@@ -231,12 +231,12 @@ func Simplify(g geom.Geometry, tolerance float64) (geom.Geometry, error) {
 	return rawgeos.Simplify(g, tolerance)
 }
 
-// TopologyPresereSimplify creates a simplified version of a geometry using
+// TopologyPreserveSimplify creates a simplified version of a geometry using
 // the Douglas-Peucker algorithm. An attempt is made to preserve topological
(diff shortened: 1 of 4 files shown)


geos/entrypoints.go
```diff
@@ -231,6 +231,15 @@ func Simplify(g geom.Geometry, tolerance float64) (geom.Geometry, error) {
 	return rawgeos.Simplify(g, tolerance)
 }
 
+// TopologyPresereSimplify creates a simplified version of a geometry using
```
albertteoh, line 234: Same with func name and use in test. 😄 

```suggestion
// TopologyPreserveSimplify creates a simplified version of a geometry using
```

albertteoh (approved):

geos/entrypoints.go
peterstace: Oops! Thanks for noticing that! (cde8f02015bc32707820e0561fa4cdda6b9971b3).

peterstace:
Thanks for reviewing @albertteoh!

PR #623, opened 2024-06-07
Add GEOS wrapper for `GEOSCoverageSimplifyVW_r`

peterstace:
## Description

See doc comment in `geos/entrypoints.go` for what the function does.

## Check List

Have you:

- Added unit tests? Yes.

## Related Issue

- N/A


Makefile
```diff
@@ -1,5 +1,5 @@
 .PHONY: all
-all: unit lint pgscan cmppg cmpgeos
+all: unit lint geos pgscan cmppg cmpgeos
```
peterstace, line 2: This was accidentally missed in a previous change, so slipping it in here.

geos/entrypoints_test.go
```diff
@@ -927,3 +935,19 @@ func TestCoverageUnion(t *testing.T) {
...
+
+func TestCoverageSimplifyVW(t *testing.T) {
+	input := geom.NewGeometryCollection(
+		[]geom.Geometry{
+			geomFromWKTFile(t, "testdata/coverage_simplify_input_birchgrove.wkt"),
+			geomFromWKTFile(t, "testdata/coverage_simplify_input_balmain.wkt"),
+		},
+	)
+	got, err := geos.CoverageSimplifyVW(input.AsGeometry(), 0.001, false)
+	if err != nil && strings.Contains(err.Error(), "unsupported") {
```
peterstace, line 947: I've used string comparison for a sentinel error, which is not ideal.

I plan to fix this using the same treatment as in https://github.com/peterstace/simplefeatures/pull/616 (there are a few instances in this package that could benefit from it, so I'll do them all at once).

peterstace: https://github.com/peterstace/simplefeatures/pull/624/files

peterstace:
Thanks Albert!

PR #624, opened 2024-06-07
Use sentinel error rather than string matching in GEOS tests

peterstace:
## Description

This makes the tests more robust to changes in the error messages, and also provides a code link back to the checked error for readers of the tests.

## Check List

Have you:

- Added unit tests? Yes.

- Updated release notes? (if appropriate?) N/A, minor internal change.

## Related Issue

- N/A


internal/rawgeos/errors.go
peterstace, line 1: Even though `unsupportedGEOSVersionError` is renamed to `UnsupportedGEOSVersionError`, this _isn't_ a change to the public interface since this package is internal only.

albertteoh (approved):
👍🏼

peterstace:
Thanks Albert!

Issue #534, opened 2023-09-21
Separate TWKB Unmarshallers


peterstace:
TWKB currently has several functions that unmarshal some aspects of TWKB:

- `UnmarshalTWKB`
- `UnmarshalTWKBWithHeaders`
- `UnmarshalTWKBBoundingBoxHeader`
- `UnmarshalTWKBEnvelope`

We don't need this many (we should be able to get away with just 2):

- `UnmarshalTWKB`
- `UnmarshalTWKBHeader`

We should also consider returning the header as a struct containing all of the header fields rather than as individual parts.

PR #627, opened 2024-06-17
Refactor TWKB unmarshal interface

peterstace:
## Description

This change modifies the TWKB unmarshal interface while minimally changing its internal implementation (I may change the internal implementation later).

There were a few problems with the previous interface that prompted this change:

- All non-geometry data was returned by a single function (`UnmarshalTWKBWithHeaders`), making the function signature awkward (with three non-error return values).

- `UnmarshalTWKBWithHeaders` also returned the geometry decoded from the TWKB, which is discordant with returning the envelope. The whole point of being able to parse the envelope is so that the user can skip full geometry decoding.

- Three different functions would provide the envelope: `UnmarshalTWKBBoundingBoxHeader`, `UnmarshalTWKBEnvelope`, and `UnmarshalTWKBWithHeaders`. There was a lot of overlap in functionality between these, and it wasn't clear to callers which one they should use.

- Some functions expose the extended envelope (an XY envelope with the addition of Z and M ranges) as a `[]Point`. This way of representing an envelope is awkward for users since nothing in the type signature enforces that there are two elements in the slice. While comments indicate that the slice will contain two elements, the type system doesn't enforce this. Furthermore, return values from the function represent empty envelopes via an empty slice. While comments communicate this behaviour, users could easily miss this.

- When the ID list is present in the TWKB, it is returned as a `[]int64`. This is error-prone for users since nothing in the type signature suggests that the slice may be empty to indicate that no ID list exists (this is only conveyed in comments).

The interface is tweaked to address the problems outlined above:

- `UnmarshalTWKB` **stays exactly the same** (and simply unmarshals the geometry part of the TWKB).

- The `UnmarshalTWKBWithHeaders`, `UnmarshalTWKBBoundingBoxHeader`, and `UnmarshalTWKBEnvelope` functions are removed.

- `UnmarshalTWKBEnvelope` is added back but with a different function signature. It now returns an `ExtendedEnvelope` type, which is a struct that contains an `Envelope` and two `Interval's (for Z and M ranges). It also returns a boolean flag. This function signature, combined with the `ExtendedEnvelope` type, explicitly indicates that the envelope is optional and that the Z and M ranges may not be present even when the envelope is.

- `UnmarshalTWKBList` is added, which returns the ID list from the TWKB (if it exists). This function returns a `[]int64` and a boolean flag indicating if the list is present in the TWKB.

- `UnmarshalTWKBSize` is added, which returns the size of the TWKB in bytes. This can be used to quickly scan through a concatenated sequence of TWKBs, just reading each TWKBs bounding box (for example).  TWKB size was not previously exposed to users, but is a useful part of the TWKB spec.

## Check List

Have you:

- Added unit tests? Yes.

## Related Issue

- https://github.com/peterstace/simplefeatures/issues/534


geom/interval.go
```diff
@@ -0,0 +1,28 @@
...
+	if boundB < boundA {
+		boundA, boundB = boundB, boundA
+	}
+	return Interval{boundA, boundB, true}
+}
+
+// MinMax returns the minimum and maximum bounds of the interval. The boolean
+// return value indicates if the interval is non-empty (the minimum and maximum
+// bounds should be ignored if false).
+func (i Interval) MinMax() (float64, float64, bool) {
```
albertteoh, line 26: Another idiom I've seen used is with `time.Time` and `uuid.UUID` (google's) which use `IsZero` and `uuid.Nil` respectively. I wonder if we should follow the same idiom?

albertteoh (approved):
👍🏼

geom/interval.go
peterstace: This is something I considered, but I don't think it's the way to go for the following reasons. I'll use `time.Time` as the point of comparison:

- `time.Time`'s zero value represents the instant "Jan 1, year 1, 00:00:00 UTC", and is detected by `IsZero()`.
- That's a valid time, albeit one that would be unlikely to need to be represented in practice.
- Mapping that over to `Interval`, the zero value would be the interval `[0, 0]` (i.e. the interval containing just the number `0`).
- But the `[0, 0]` interval would be useful in _a lot_ of contexts that aren't rare in the same way that "Jan 1, year 1, midnight" is rare. For example, as an `ExtendedEnvelope`, the interval might indicate a range of elevations that is perfectly flat (i.e. all zero height compared to some reference height). That needs to be differentiated from an `ExtendedEnvelope` that doesn't have any elevation data.
- The other approach could be to keep the `nonEmpty` boolean inside `Interval`, but change the signature of `MinMax` to `MinMax() (float64, float64)` and add an `IsZero() bool` method (which might have been what you were thinking?). If `IsZero()` returns `true`, then the return value of `MinMax()` should be ignored. That would represent everything ok, but I'm really not a fan of those types of APIs, since it's very easy for users to miss that the type represents a potentially empty interval (i.e. they just see the `MinMax` method and ignore the `IsZero` method, not realising that `Interval` may represent an empty interval).

albertteoh: Good points, and agree!

peterstace:
Thanks for reviewing Albert, I really appreciate it!

PR #628, opened 2024-06-19
Update docker compose setup for best practices

peterstace:
## Description

This change makes several cosmetic/non-function changes to how docker compose is set up:

- Files in the form `docker-compose-xxx.yml` are renamed to `compose-xxx.yaml` to better follow the normal naming practice for compose files. See https://docs.docker.com/compose/compose-application-model/#the-compose-file.  Since docker compose V2, the `docker-` prefix is no longer used.

- Removes the `version` field from the compose files. This field is obsolete, and its presence was causing a warning to be printed. See https://github.com/compose-spec/compose-spec/blob/master/spec.md#version-and-name-top-level-elements

- Explicitly sets `tags` as an empty string to remove a warning about it not being set.

## Check List

Have you:

- Added unit tests? N/A

## Related Issue

- N/A


albertteoh (approved):
👍🏼

peterstace:
Thanks Albert 😄

Issue #631, opened 2024-08-10
Polygon Validation Error


Amalio769:
With a wrong Polygon (the internal ring is bigger that external one, and the internal one intersect with the external one in more than one point) the function Validate() work properly.

But if the internal ring intersect in only one point with the external ring, the function Validate don't work properly.
Please, check this Polygon with the following coordinates: [[[1,1],[5,1],[5,5],[1,5],[1,1]],[[3,1],[6,0],[6,6],[0,6],[0,0],[3,1]]]

peterstace:
Hi @Amalio769 , thank you for your report. I've confirmed that the geometry is invalid via QGIS, and I've been able to reproduce the bug in simplefeatures where it's erroneously indicating that it's valid. I'll look into a fix (I've got an intuitive understanding of where I think the problem most likely is).

peterstace:
I've got the bug reproduction in this branch: https://github.com/peterstace/simplefeatures/compare/master...polygon_validation_bug

Amalio769:
I've compile my repo with the new brach and all my tests work properly. Good job. Thank you very much.

peterstace:
FYI the fix is in the v0.51.0 release. https://github.com/peterstace/simplefeatures/releases/tag/v0.51.0

PR #632, opened 2024-08-14
Fix bug in polygon validation

peterstace:
## Description

The bug caused some invalid polygons to be incorrectly identified as valid. Specifically, the bug occurred when an inner ring is outside of the outer ring, and the first control point of the inner ring is touching the outer ring.

## Check List

Have you:

- Added unit tests? Yes.

## Related Issue

- https://github.com/peterstace/simplefeatures/issues/631

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## Unreleased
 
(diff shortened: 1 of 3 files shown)


albertteoh (approved):
👍🏼

peterstace:
Thanks for the review @albertteoh ! 🥳

PR #635, opened 2024-08-29
Upgrade to golangci-lint v0.60.3

peterstace:
## Description

The only breakages were related to `gosec`, with the "G115" rule being triggered (the rule relates to potential integer overflows). These were all false positives, so I just disabled that particular rule.

## Check List

Have you:

- Added unit tests? N/A

## Related Issue

- N/A


albertteoh (approved):

peterstace:
Thanks for reviewing!

Issue #595, opened 2024-03-12
Function for NewCoordinates


akhenakh:
I run into a stupid bug because specifying Z in a `NewPoint(geom.Coordinates{})` results in it defaulting to `Type: geom.DimXY`, it makes sense.

```go
package main

import (
	"fmt"

	"github.com/peterstace/simplefeatures/geom"
)

func main() {
	p := geom.NewPoint(geom.Coordinates{
		XY: geom.XY{
			X: 10,
			Y: 11,
		},
		Z: 400,
	}).AsGeometry()

	fmt.Println(p.DumpCoordinates())
	fmt.Println(p.AsText())
}
```

In the rest of the API, though,  helper functions helps to remind you to specify the dimension like `geom.NewSequence(seq, geom.DimXY)`.

I did not find any `NewCoordinates()` ?

peterstace:
Yeah I see what you mean, thank you for reporting your experience. I think there are a few different options to make this sort of bug less likely:

# Option 1

Add a `geom.NewCoordinates` function that takes a `geom.CoordinatesType`. We'd also have to make it accept the right number of float64 arguments. So maybe the function signature would be like `geom.NewCoordinates(geom.CoordinatesType, xyzm ...float64)`. Examples of use:

- `geom.NewPoint(geom.NewCoordinates(geom.DimXY, 10, 11))`
- `geom.NewPoint(geom.NewCoordinates(geom.DimXYZ, 10, 11, 400))`

It might be best if it panics if the `geom.CoordinatesType` argument doesn't match the number of args, which is the same as what `geom.NewSequence` does.

# Option 2

Create new functions for each coordinates type:

- `geom.NewCoordinatesXY(x, y float64) geom.Coordinates`
- `geom.NewCoordinatesXYZ(x, y, z float64) geom.Coordinates`
- `geom.NewCoordinatesXYM(x, y, m float64) geom.Coordinates`
- `geom.NewCoordinatesXYZM(x, y, z, m float64) geom.Coordinates`

Usage examples:

- `geom.NewPoint(geom.NewCoordinatesXY(10, 11))`
- `geom.NewPoint(geom.NewCoordinatesXYZ(10, 11, 400))`

# Option 3

The `geom.Coordinates` are mainly used for the creation of `geom.Point` values. So maybe we could just create new point constructors like this:


- `geom.NewPointXY(x, y float64) geom.Point`
- `geom.NewPointXYZ(x, y, z float64) geom.Point`
[...]

Issue #477, opened 2022-11-11
Make polygon construction easier


peterstace:
If a user has some input coordinates, it is quite onerous to get a polygon object.

For example, the following program has a hardcoded list of coordinates, constructs a polygon from those coordinates, and then calculates the centroid of the polygon.

```
func main() {
    coords := []float64{
        144.9324481856214,
        -37.80831791530031,
        151.1946474301501,
        -33.85326738062629,
        153.011491595451,
        -27.465643838083587,
        130.84533948192757,
        -12.45094813855576,
        115.8443383313529,
        -31.96739822401411,
        144.9324481856214,
        -37.80831791530031,
    }
    seq := geom.NewSequence(coords, geom.DimXY)
    ring, err := geom.NewLineString(seq)
    if err != nil {
        log.Fatalf("could not build ring: %v", err)
    }
    poly, err := geom.NewPolygon([]geom.LineString{ring})
    if err != nil {
        log.Fatalf("could not build polygon: %v", err)
    }

    log.Printf("Center: %v", poly.Centroid().AsText())
}
```

It would be nice if there were alternate (easier/terser) ways to create geometry objects. Potentially this could live in a separate package (named `gbuild` in this example, but that's just a placeholder). For example, it might look like this:

```
func main() {
    coords := []float64{
        144.9324481856214,
        -37.80831791530031,
        151.1946474301501,
        -33.85326738062629,
        153.011491595451,
        -27.465643838083587,
        130.84533948192757,
        -12.45094813855576,
        115.8443383313529,
        -31.96739822401411,
        144.9324481856214,
        -37.80831791530031,
    }
    poly, err := gbuild.PolygonFromCoords([][]float64{coords})
        log.Fatalf("could not build polygon: %v", err)
    }

    log.Printf("Center: %v", poly.Centroid().AsText())
}
```

peterstace:
Lot's of "helpers" could live in this new package. Some ideas are:

- `Must` variants that panic on error. The user doesn't need to handle errors for these if they're sure the inputs are valid (or are willing to risk a panic).
- Constructors for each geometry type directly from `float64` coordinates.
- XYZ, XYM, and XYZM are out of scope. These have a pretty neiche usecases (and the regular `[]float64` -> `Sequence` -> `Geoometry` path is always available).
- Include `Envelope` constructors as well.
- The need to create regular polygons (approximating circles) tends to come up a lot. That may be useful to add too.

peterstace:
[...]

PR #636, opened 2024-09-27
Geometry constructors that accept coordinates

peterstace:
## Motivation

Constructing geometries _not_ from WKTs, WKBs, TWKBs, or GeoJSON is currently
very arduous. E.g, this is how you would construct a `Polygon` from a list of
coordinates:

```go
coords := []float64{
    144.9324481856,
    -37.8083179153,
    151.1946474301,
    -33.8532673806,
    153.0114915954,
    -27.4656438380,
    130.8453394819,
    -12.4509481385,
    115.8443383313,
    -31.9673982240,
    144.9324481856,
    -37.8083179153,
}
seq := geom.NewSequence(coords, geom.DimXY)
ring := geom.NewLineString(seq)
poly := geom.NewPolygon([]geom.LineString{ring})
```

This is annoying in particular for users of simplefeatures who wish to
construct geometries programmatically (e.g. for their own unit tests, or other
"hardcoded" geometries).

## Mechanism

This change adds a new group of functions to the `geom` package that construct
geometries, in an easier way than the existing `Sequence` -> `LineString` ->
`Polygon` -> `MultiPolygon` etc. chain.

These functions have names in the form `New{Type}{Dimension}`, where `Type` is
one of {`Point`, `LineString`, `Polygon`, `MultiPoint`, `MultiLineString`,
`MultiPolygon`}, and `Dimension` is one of {`XY`, `XYZ`, `XYM`, `XYZM`}.

They accept coordinates at the appropriate level of nesting that corresponds to
the geometry type. For example, `NewPolygonXY` accepts a variadic number of
slices of `float64` (each slice represents a ring). `NewMultiPolygonXY` accepts
a variadic number of slices of slices of `float64` (each outer slice represents
a polygon, and each inner slice represents a ring).

These functions panic if the number of coordinates provided is not consistent
with the dimension (this behaviour is consistent with the behaviour of the
`NewSequence` function).

The example above can no be written as:

```go
coords := []float64{
    144.9324481856,
    -37.8083179153,
    151.1946474301,
    -33.8532673806,
    153.0114915954,
    -27.4656438380,
    130.8453394819,
    -12.4509481385,
    115.8443383313,
    -31.9673982240,
    144.9324481856,
    -37.8083179153,
}
poly := geom.NewPolygonXY(coords)
```

## Alternatives considered

The main alternative I considered was not having `Dimension` suffix variants,
and instead accepting a `geom.CoordinatesType` value as the first argument
(which can take one of the values {`geom.DimXY`, `geom.DimXYZ`, `geom.DimXYM`,
`geom.DimXYZM`}). This would have the advantage that the number of functions is
reduced by a factor of 4, but introduces some other disadvantages:

- The length of the invocation is longer at the callsite. E.g.
  `geom.NewPolygonFromCoords(geom.DimXY, coords)` vs
  `geom.NewPolygonXY(coords)` (the `geom.NewPolygon` function is already taken,
  so the `FromCoords` suffix is necessary).

- The `NewPointFromCoords` function would have suboptimal ergonomics, as it
  would accept a `geom.CoordinatesType` value and a variadic number of
  `float64` values. Typically, the `geom.CoordinatesType` would be hardcoded,
  so the dimensionality is encoded twice even though it doesn\'t have to be
  (once in the `geom.CoordinatesType`, and once in the number of `float64`
  values). Alternatively, the `geom.CoordinatesType` could be omitted, but then
  `NewPointFromCoords` would be inconsistent with the other `NewXXXFromCoords`
  functions.

## Check List

Have you:

- Added unit tests? Yes.

## Related Issue

- https://github.com/peterstace/simplefeatures/issues/477
- https://github.com/peterstace/simplefeatures/issues/595

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,15 @@
 
 - Upgrades `golangci-lint` to `v1.60.3`.
 
+- Adds a new group of functions to the `geom` package that construct
+  geometries. These functions have names in the form `New{Type}{Dimension}`,
+  where `Type` is one of {`Point`, `LineString`, `Polygon`, `MultiPoint`,
+  `MultiLineString`, `MultiPolygon`}, and `Dimension` is one of {`XY`, `XYZ`,
+  `XYM`, `XYZM`}. They accept coordinates at the appropriate level of nesting
+  that corresponds to the geometry type. These functions panic if the number of
+  coordinates provided is not consistent with the dimension (this behaviour is
+  consistent with the behaviour of the `NewSequence` function).
+
 ## v0.51.0
 
 2024-08-19
--- a/geom/ctor_from_coords.go
+++ b/geom/ctor_from_coords.go
@@ -309,30 +309,31 @@ func multiPointFromCoords(coords []float64, ct CoordinatesType) MultiPoint {
 		return MultiPoint{}.ForceCoordinatesType(ct)
 	}
 
-	if len(coords)%ct.Dimension() != 0 {
+	dim := ct.Dimension()
+	if len(coords)%dim != 0 {
 		msg := fmt.Sprintf(
 			"geom: coordinate arguments to %s constructor "+
 				"must have a length that is a multiple of %d",
-			ct.String(), ct.Dimension(),
+			ct.String(), dim,
 		)
 		panic(msg)
 	}
 
-	n := len(coords) / ct.Dimension()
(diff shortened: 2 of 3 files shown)


albertteoh (approved):
Looks good!

peterstace:
Thanks for reviewing @albertteoh , greatly appreciated!

PR #652, opened 2025-05-23
Add PROJ wrapper

peterstace:
## Description

The wrapper exists in the `github.com/peterstace/simplefeatures/proj` package. It leverages the PROJ library to allow geometries to be transformed between a vast number of coordinate reference systems.

## Check List

Have you:

- Added unit tests? Yes.


internal/test/test.go
peterstace, line 1: I've noticed that test helpers are starting to get copied/pasted everywhere... So I think it's time to make a new central package to hold them (i.e. this one). I may go over the whole module at a later point in time and try to properly consolidate the test helpers.

Issue #650, opened 2025-04-20
Support for DimXYZM in GeoJSON unmarshaling


GruffGemini:
Hello,

I have noticed that when unmarshaling a GeoJSON, coordinates with more than 3 dimensions are allowed, but ignored, as seen here
https://github.com/peterstace/simplefeatures/blob/17a6536281b9c96c2f42e030b6e78d1ff8ba5ccc/geom/geojson_unmarshal.go#L57C2-L60C3

Since this package supports XYZM dimension, is it possible to add support for unmarshaling such GeoJSONs (at least for the case when all coordinates in a GeoJSON have strictly 4 dimensions, so there is no ambiguity)?

peterstace:
Hi,

The RFC 7946 has [a section](https://datatracker.ietf.org/doc/html/rfc7946#section-3.1) related to this:

> Implementations SHOULD NOT extend positions beyond three elements because the semantics of extra elements are unspecified and ambiguous.  Historically, some implementations have used a fourth   element to carry a linear referencing measure (sometimes denoted as   "M") or a numerical timestamp, but in most situations a parser will  not be able to properly interpret these values.  The interpretation   and meaning of additional elements is beyond the scope of this   specification, and additional elements MAY be ignored by parsers.

At first glance, the spec seems to be a bit ambiguous...

- `Implementations SHOULD NOT extend positions beyond three elements` seems clear (i.e. don't support XYZM).
- But `additional elements MAY be ignored by parsers` seems to conflict with that... If additional elements "MAY" be ignored by parsers, that seems to indicate that they don't _need_ to be ignored.

But maybe bullet point one is talking about marshalling, and bullet point two is talking about unmarshalling. In that case, it's probably ok to do as you suggest.

PR #658, opened 2025-06-20
Support XYZM in GeoJSON unmarshalling

peterstace:
## Description

The GeoJSON spec states:

> Implementations SHOULD NOT extend positions beyond three elements because the
> semantics of extra elements are unspecified and ambiguous. Historically, some
> implementations have used a fourth element to carry a linear referencing
> measure (sometimes denoted as "M") or a numerical timestamp, but in most
> situations a parser will not be able to properly interpret these values. The
> interpretation and meaning of additional elements is beyond the scope of this
> specification, and additional elements MAY be ignored by parsers.

In particular `additional elements MAY be ignored by parsers` indicates that parsers may (or may not, in our case) choose to interpret additional elements.

This changes interprets the 4th element as a measure (M) value.

Marshalling behaviour is unchanged, so the 4th element will not be included in the marshalled GeoJSON output. This is required as per the first sentence in the quote above.

## Check List

Have you:

- Added unit tests? Yes.

## Related Issue

- https://github.com/peterstace/simplefeatures/issues/650

### DIFF
--- a/geom/geojson_unmarshal.go
+++ b/geom/geojson_unmarshal.go
@@ -19,7 +19,7 @@ func UnmarshalGeoJSON(input []byte, nv ...NoValidate) (Geometry, error) {
 		return Geometry{}, err
 	}
 
-	hasLength := make(map[int]bool)
+	hasLength := make(map[int]struct{})
 	if err := detectCoordinatesLengths(rootObj, hasLength); err != nil {
 		return Geometry{}, err
 	}
@@ -34,7 +34,7 @@ func UnmarshalGeoJSON(input []byte, nv ...NoValidate) (Geometry, error) {
 	return g, nil
 }
 
-func chooseGeoJSONCoordinatesType(hasLength map[int]bool) CoordinatesType {
+func chooseGeoJSONCoordinatesType(hasLength map[int]struct{}) CoordinatesType {
 	// The GeoJSON spec allows (but doesn't require) parsers to ignore any
 	// "extra" coordinate values in addition to the normal 3 coordinate values
 	// used to specify a 3D position. We choose to interpret a 4th dimension as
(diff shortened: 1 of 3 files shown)


geom/geojson_unmarshal.go
```diff
@@ -68,6 +34,42 @@ func UnmarshalGeoJSON(input []byte, nv ...NoValidate) (Geometry, error) {
...
+	// | false | false | false      | DimXY   |
+	// | false | false | true       | DimXYZM |
+	// | false | true  | false      | DimXYZ  |
+	// | false | true  | true       | DimXYZ  |
+	// | true  | false | false      | DimXY   |
+	// | true  | false | true       | DimXY   |
+	// | true  | true  | false      | DimXY   |
+	// | true  | true  | true       | DimXY   |
+	var has2, has3, has4OrMore bool
+	for length := range hasLength {
```
peterstace, line 59: `hasLength` is a "set", but we're not checking the bool here. Would be better to use a `map[int]struct{}` instead to make it less ambiguous.

geom/geojson_unmarshal.go
```diff
@@ -324,9 +326,23 @@ func oneDimFloat64sToCoordinates(fs []float64, ctype CoordinatesType) (Coordinat
...
 		Type: ctype,
 	}
-	if ctype.Is3D() {
+
+	if len(fs) < ctype.Dimension() {
+		// This should not happen because we used the lengths to choose the ctype.
+		panic(fmt.Sprintf("unexpected length %d for ctype %s", len(fs), ctype))
+	}
+
+	switch ctype {
```
peterstace, line 335: Should have a `DimXY` case that does nothing, and a `default` that panics.

PR #674, opened 2025-11-07
DRAFT/WIP: Make rtree package generic over record type

peterstace:
## Description

The `rtree` package previously hardcoded record IDs as integers, requiring users to maintain separate mappings between their records and integer IDs. This added unnecessary indirection and made the API less ergonomic.

By making `RTree` generic over the record type T, users can now store their actual records directly in the tree (or use integers if they prefer to keep the existing behaviour). This simplifies usage patterns and eliminates the need for separate index mappings in consuming code.

Note: This is a breaking change. The easiest way for users to upgrade is to simply replace all occurrences of `*rtree.RTree` with `*rtree.RTree[int]` in their codebase.

## Check List

Have you:

- Added unit tests? Yes.

### DIFF
--- a/geom/rtree.go
+++ b/geom/rtree.go
@@ -3,7 +3,7 @@ package geom
 import "github.com/peterstace/simplefeatures/rtree"
 
 // TODO: Use this instead of indexedLines/Points where possible.
-func newLineRTree(lines []line) *rtree.RTree[line] {
+func newLineRTree(lines []line) *rtree.RTree[line] { //nolint:unused
 	items := make([]rtree.BulkItem[line], len(lines))
 	for i, ln := range lines {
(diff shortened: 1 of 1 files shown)


.golangci.yaml
```diff
@@ -35,8 +35,6 @@ linters-settings:
   errcheck:
     exclude-functions:
       - io.Copy(os.Stdout)
-      - (*github.com/peterstace/simplefeatures/rtree.RTree).RangeSearch
```
peterstace: `errcheck` seems to get a bit confused when generics are introduced... I can longer specify funcs in `exclude-functions` section. Instead, I've opted to just use `_ = ...` whenever I want to ignore an error.

